PR #9, opened 2017-08-18
modify poa protobuf block and  Normal block

shengling123:


consensus/authority_round/src/core/authority_round.rs
```diff
@@ -19,21 +19,20 @@ use super::{Engine, EngineError, Signable, unix_now, AsMillis};
...
-use util::H256 as Hash256;
 use std::sync::{Arc, Mutex};
 use std::sync::atomic::{AtomicUsize, Ordering, AtomicBool};
 use std::sync::mpsc::Sender;
 use std::time::Duration;
 use tx_pool::Pool;
 use util::Address;
-use util::Hashable;
+
+use util::H256 as Hash256;
```
u2: 可以删掉 `use util::H256 as Hash256;`

shengling123: OK

PR #17, opened 2017-08-21
Improve txtool; add missing dependency.

jerry-sl:
Improve txtool; add missing dependency.


u2:
add missing dependency ??

jerry-sl:
@u2  Requests package.

PR #37, opened 2017-08-30
task support solidity compatible storage

urugang:
implement types such as U256/String/Array/Map, see test for example  
improve contract test using solc  
remove unnecessary plugin annotation


u2:
这个写个性能测试吧，对比下EVM的执行？ @vinberm 来试下？

vinberm:
OK @u2

PR #58, opened 2017-09-05
send chain.status to network.

jerry-sl:
Still need to send chain.status to network when Chain is up and add_block.


u2:
Chain 里面的 Status 可以删掉了吧？

PR #63, opened 2017-09-06
Fixed contract address in receipt

u2:
contractAddress: DATA, 20 Bytes - The contract address created, if the transaction was a contract creation, otherwise null.


rink1969:
但是多了这段，感觉也不会有影响啊

u2:
是没有影响，但是这个contractAddress表示的就是创建的合约地址。

PR #117, opened 2017-09-16
check account gas limit for all and special account

u2:


chain/core/src/libchain/block.rs
```diff
@@ -281,16 +283,22 @@ impl DerefMut for OpenBlock {
...
         let r = OpenBlock {
             exec_block: ExecutedBlock::new(block, state, tracing),
             last_hashes: last_hashes,
             tx_hashes: Vec::new(),
+            account_gas_limit: 10000.into(), // TODO: account_gas_limit
+            account_gas: special_gas_limit.iter()
+                                          .map(|(key, value)| (key, min(value, &account_gas_limit)))
+                                          .fold(HashMap::new(), |mut acc, (key, value)| {
+                acc.insert(*key, *value);
+                acc
```
u2: 这里是干嘛的？

PR #122, opened 2017-09-18
add error processing for send_data

hezhengjun:
add the error processing branch for send_data to avoid panic


rink1969:
已经跟 #120 一起合入了

PR #135, opened 2017-09-22
prune backing and journal after commit block

urugang:
1) prune backing and journal after commit block, keep latest 2 history block
2) clean unused crates


u2:
stable_holiday 也需要提交PR

PR #144, opened 2017-11-14
Merge dev_with_zcash_senlody

Senlody:
New zk-proof strategies using homomorphic encryption
With round_test(a test of one integrate trade)


janx:
@Senlody thanks :beers:

PR #152, opened 2018-02-10
Refactor: remove unfriendly comment

suyanlong:


cita-network/src/syncprotol.md
```diff
@@ -1,71 +0,0 @@
-# 同步协议
-
-关键字解释:
```
u2: why remove this doc?

u2: Chinese comment is better than none.

PR #174, opened 2018-08-04
Update getting_started.md

jasl:
Fix a layout issue


jasl:
I found this tiny bug when I'm reading <https://docs.nervos.org/cita/#/./chain/getting_started> so just click "edit" then do this quick fix.

As your suggest, I clone Cita to trying fix English version, but I find big problems:

First, [There have `Latest` and `latest` in one directory](https://github.com/cryptape/cita/tree/gh-pages/zh-CN), macOS's filesystems case-insensitive by default, so I can only see one of them.

Second, <https://docs.nervos.org/cita/#/./chain/getting_started> using [`Latest`](https://github.com/cryptape/cita/blob/gh-pages/zh-CN/Latest/chain/getting_started.md) and difference with `latest` one, that make me confusion.

Third, I roughly read English version of `Latest` and `latest`, they look the same, so English version doesn't have this bug.

Fourth, English version have problem on GH page.

image

kaikai1024:
Great! We will fix it soon.

And use the `Latest` directory. You can check the same path on develop branch, and you'll see.The `latest` dir should be deleted, there may be some wrong when updating the code cause we work on bitbucket. 

thx for that

zh-CN/Latest/chain/getting_started.md
```diff
@@ -284,4 +284,4 @@ $ ./env.sh ./bin/cita help
 > 在发布件目录(target/install)下运行节点。
 > 可选择使用`./env.sh`
 
-更多 API（如合约调用、交易查询）请参见[RPC 调用](./rpc_guide/rpc)。
\ No newline at end of file
+更多 API（如合约调用、交易查询）请参见[RPC 调用](./rpc_guide/rpc)。
```
kaikai1024: ? what's the difference? EOF?

jasl: Maybe, I don't edit this line.

PR #217, opened 2019-01-31
Feature/upgrade bft

keroro520:
# Pull Request Template

### Requirements
- Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
- All new codes require tests to ensure against regressions.

Old snapshot_tool has too many implicit conventions. That introduces many complexity to snapshot, and makes debug harder, and even brings Bugs.
So we have rewritten snapshot_tool to require responses for all kinds of snapshot-requests. So here cita-bft upgrade to do that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 CARGO=RUSTFLAGS='-F warnings' cargo
 
+.PHONY: debug release test test-release bench fmt cov clean clippy security_audit
+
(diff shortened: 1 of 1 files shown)


kaikai1024:
The circle ci is coming...
Please be patient for the travis ci

PR #255, opened 2019-02-24
fix(network): Try reparsing the socket address

yejiayu:
# Pull Request Template

### Requirements
ref #250 

When parsing the socket address, it will retry if it fails.

Note: There is still a problem updating the `network` configuration

@jiangxianliang007 Please test this branch on the k8s cluster.

### DIFF
--- a/cita-network/src/connection.rs
+++ b/cita-network/src/connection.rs
@@ -219,11 +219,13 @@ impl Connections {
             Manager::new(connect_task_sender, connect_receiver, enable_tls).run()
         });
 
-        if let Some(peers) = config.peers.as_ref() {
-            config_peer(&peers).into_iter().for_each(|config| {
-                connect_sender.send(config).unwrap();
-            });
-        }
+        let init_connect_sender = connect_sender.clone();
+        let init_config = config.clone();
+        thread::spawn(move || {
+            if let Some(peers) = init_config.peers.clone() {
+                init_config_peer(peers, init_connect_sender)
+            }
+        });
 
         (
             Connections {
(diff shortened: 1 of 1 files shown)


yejiayu:
PTAL @kaikai1024 @leeyr338

yejiayu:
Is it normal for ci to fail?

u2:
> Is it normal for ci to fail?

No. Check format.

ouwenkg:
Checking clippy failed.

yangby-cryptape:
> error[E0308]: mismatched types
>    --> cita-network/src/connection.rs:226:41

Have you built the project before pulling request? This is not a lint error!

CI is really important for this situation.

yangby-cryptape:
> Is it normal for ci to fail?

Have you checked the CI logs before you asked?

It is OK if you just confused with the CI logs, ask and someone will answer it, but, if you just are lazy ...
Come on, u are a grown-up!

yejiayu:
> Have you checked the CI logs before you asked?
> 
> It is OK if you just confused with the CI logs, ask and someone will answer it, but, if you just are lazy ...
> Come on, u are a grown-up!

I checked the previous pr and found that many ci failed, so I raised this question.

kaikai1024:
> > Have you checked the CI logs before you asked?
> > It is OK if you just confused with the CI logs, ask and someone will answer it, but, if you just are lazy ...
> > Come on, u are a grown-up!
> 
> I checked the previous PR and found that many ci failed, so I raised this question.

Because the circleCI was not ready that time.

cita-network/src/connection.rs
```diff
@@ -427,7 +429,7 @@ pub fn manage_connect(config_path: &str, rx: Receiver<DebouncedEvent>, task_send
 
 fn config_peer(config: &[PeerConfig]) -> Vec<(u32, SocketAddr, String)> {
```
kaikai1024: Use `init_config_peer` instead of `config_peer`? Then why change this function?

yejiayu: The update configuration also depends on "config_peer".

jiangxianliang007:
Tested through the K8S cluster

kaikai1024:
> Tested through the K8S cluster

We'll no release a patch version, That's right?

Issue #229, opened 2019-02-14
body of getBlock don't return from address in v0.21rc


classicalliu:
## Description

`getTransaction` returns `from` address in v0.21rc, but body of `getBlockByNumber` does not.

## Versions

v0.21rc

## Additional Information

`getTransaction`

![image](https://user-images.githubusercontent.com/13375784/52770388-e864cf00-306d-11e9-8c95-f111c3c4c42f.png)

`getBlockByNumber`

![image](https://user-images.githubusercontent.com/13375784/52770412-f74b8180-306d-11e9-828b-6ce482b18355.png)

kaikai1024:
I will check that ASAP.

kaikai1024:
@classicalliu 

You can try to resolve this issue by yourself. you can refer to the [commit](https://github.com/cryptape/cita-common/commit/f85faf2bd8169c57e143b00b500e90a4fa3e80c3)

classicalliu:
OK, I'll try

kaikai1024:
> OK, I'll try

Cool

kaikai1024:
https://github.com/cryptape/cita-common/commit/f85faf2bd8169c57e143b00b500e90a4fa3e80c3

kaikai1024:
You can continue to update cita-common submodule. @classicalliu

PR #258, opened 2019-02-25
Update cita-common and bft for add `from` to `body` of `getBlockByNumber`

classicalliu:
Add `from` to `body` of `getBlockByNumber` and `getBlockByHash`

image

#### Applicable Issues
ref #229


kaikai1024:
I will check the travis later.

PR #264, opened 2019-02-26
Feature/network p2p

leeyr338:
Use crossbeam.


kaikai1024:
You need to ensure the commites are clean when you try to merge it to the develop.

Now it's ok for me.

PR #265, opened 2019-02-27
Bugfix for node cannot establish in k8s: Multi translate config address

leeyr338:
Bugfix for node cannot establish in k8s: Multi translate config address

### DIFF
--- a/cita-network/src/node_manager.rs
+++ b/cita-network/src/node_manager.rs
@@ -34,7 +34,7 @@ use std::{
     time::{Duration, Instant},
 };
 
-pub const DEFAULT_MAX_CONNECTS: usize = 4;
+pub const DEFAULT_MAX_CONNECTS: usize = 666;
 pub const DEFAULT_PORT: usize = 4000;
 pub const CHECK_CONNECTED_NODES: Duration = Duration::from_secs(3);
 type IsTranslated = bool;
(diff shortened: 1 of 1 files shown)


kaikai1024:
You can merge to your branch yourself.
And you don't need to clean the commits when finish it. `squash and merge` can be a choice after review

cita-network/src/node_manager.rs
```diff
@@ -34,13 +34,15 @@ use std::{
     time::{Duration, Instant},
 };
 
-pub const DEFAULT_MAX_CONNECTS: usize = 4;
```
kaikai1024: 666?

cita-network/src/node_manager.rs
```diff
@@ -34,13 +34,15 @@ use std::{
...
+pub const DEFAULT_MAX_CONNECTS: usize = 666;
```
kaikai1024: 666?

leeyr338: I like 666.

u2:
The feature/network-p2 branch is a private branch for yourself or a public branch for more users?

kaikai1024:
> You can merge to your branch yourself.
> And you don't need to clean the commits when finish it. `squash and merge` can be a choice after review

Also need to clean the commits. you should treat your every commits seriously.

kaikai1024:
> The feature/network-p2 branch is a private branch for yourself or a public branch for more users?
> 
> If it's just a private branch, pushing to your local fork repo is ok. Reorg the commits while PR.

we will discuss this.

cita-network/src/node_manager.rs
kaikai1024: ???? Just that? Is it ok for whatevre num?

PR #278, opened 2019-03-05
Feature/network: Support p2p for CITA

leeyr338:
Rewrite the module `network` for supporting p2p for CITA.

#### Benefits
P2P discovery, do not need  a `static config file` to config the network.

#### Verification Process
- Logic compatible with older versions, pass all the test cases.
- Some new test cases for  P2P discovery.

#### Applicable Issues
#250 #211 #251


cita-network/Cargo.toml.bak
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "cita-network"
+version = "0.6.0"
+authors = ["Cryptape Technologies <contact@cryptape.com>"]
+edition = "2018"
+
+[dependencies]
+fnv = "1.0.6"
+discovery = { git = "https://github.com/nervosnetwork/p2p.git" }
```
kaikai1024: Can we use the version?
> tentacle = "0.1"

leeyr338: Can not for now, maybe later.

cita-network/src/main.rs
```diff
@@ -140,35 +125,47 @@ fn main() {
         .author("Cryptape")
         .about("CITA Block Chain Node powered by Rust")
         .args_from_usage("-c, --config=[FILE] 'Sets a custom config file'")
+        .args_from_usage("-a, --address=[FILE] 'Sets a address file'")
```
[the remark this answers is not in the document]
leeyr338: Sure, it can!

CipherWang:
As far as I know, P2P feature has two separate main functions: 1. node discovery; 2. message relay (reducing n*n full connection). 
From system management's point of view, node discovery is very useful but not a direct upgrade of node list config file mode. For some circumstance, such as access-limited consortium chain, node automatic discovery means we cannot keep unauthorized node outside.
So I suggest that the we keep node discovery and node list mechanism at the same time (use config bit to switch), while make message relay function always work. For consortium chains, users could use node list file to achieve manual network topology setup. For public permissioned chains, users could use node discovery mechanism to add new nodes to the network automatically.

kaikai1024:
> As far as I know, P2P feature has two separate main functions: 1. node discovery; 2. message relay (reducing n*n full connection).
> From system management's point of view, node discovery is very useful but not a direct upgrade of node list config file mode. For some circumstance, such as access-limited consortium chain, node automatic discovery means we cannot keep unauthorized node outside.
> So I suggest that the we keep node discovery and node list mechanism at the same time (use config bit to switch), while make message relay function always work. For consortium chains, users could use node list file to achieve manual network topology setup. For public permissioned chains, users could use node discovery mechanism to add new nodes to the network automatically.

that's a good point. I'm not sure about the workload, It seems that @leeyr338 has refactored all the network

kaikai1024:
1. A standard mq module
2. rename `mq_sender`
3. the structure diagram

leeyr338:
> > As far as I know, P2P feature has two separate main functions: 1. node discovery; 2. message relay (reducing n*n full connection).
> > From system management's point of view, node discovery is very useful but not a direct upgrade of node list config file mode. For some circumstance, such as access-limited consortium chain, node automatic discovery means we cannot keep unauthorized node outside.
> > So I suggest that the we keep node discovery and node list mechanism at the same time (use config bit to switch), while make message relay function always work. For consortium chains, users could use node list file to achieve manual network topology setup. For public permissioned chains, users could use node discovery mechanism to add new nodes to the network automatically.
> 
> that's a good point. I'm not sure about the workload, It seems that @leeyr338 has refactored all the network

From an implementation perspective, it is easy to get this feature. But I don't think it is safe enough for consortium chain while a node is exposing in public internet. Maybe using firewall to protect the node from free access would be a common option.

cita-network/src/main.rs
```diff
@@ -140,35 +125,47 @@ fn main() {
...
+    debug!("network config is {:?}", config);
+    // <<<< End init config
+
+    let addr_path = matches.value_of("address").unwrap_or("address");
+    let own_addr = AddressConfig::new(&addr_path);
+    debug!("node address is {:?}", own_addr.addr);
+
+    // >>>> Init pubsub
+    // New transactions use a special channel, all new transactions come from:
+    // JsonRpc -> Auth -> Network,
```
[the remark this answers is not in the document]
leeyr338: Fixed

Makefile
```diff
@@ -1,5 +1,4 @@
 CARGO=RUSTFLAGS='-F warnings' cargo
-
```
kaikai1024: No need to change Makefile.

cita-network/src/node_manager.rs
```diff
@@ -193,7 +193,7 @@ impl Default for NodesManager {
             nodes_manager_service_receiver: rx,
             service_ctrl: None,
             peer_key: Address::zero(),
-            enable_tls: false,
+            enable_tls: true,
```
kaikai1024: Why change it to true?

cita-network/Cargo.toml
```diff
@@ -2,32 +2,34 @@
...
-serde = "1.0"
-serde_derive = "1.0"
-futures = "0.1"
-byteorder = "1.0.0"
-dotenv = "0.13.0"
-bytes = "0.4"
-notify = "4.0.0"
-libproto = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
+fnv = "1.0.6"
+discovery = { git = "https://github.com/nervosnetwork/p2p.git" }
```
yangby-cryptape: Why do not use [crates.io](https://crates.io/)?
At least, add a branch name, a commit-id or something else (more detail can be found in cargo docs).

cita-network/src/main.rs
```diff
@@ -134,127 +108,61 @@ fn main() {
...
-
-    // split new_tx with other msg
-    let (ctx_sub_tx, crx_sub_tx) = channel::unbounded();
-    let (ctx_pub_tx, crx_pub_tx) = channel::unbounded();
-    start_pubsub(
-        "network_tx",
-        routing_key!([Auth >> Request, Auth >> GetBlockTxn, Auth >> BlockTxn]),
-        ctx_sub_tx,
-        crx_pub_tx,
+    // >>>> Init config
```
yangby-cryptape: A bit confusion about this docs syntax.

cita-network/Cargo.toml
leeyr338: Discovery has not released yet.

As you said, a branch name and commit-id should be added.

cita-network/src/main.rs
leeyr338: Ok, I will delete this.

cita-network/src/node_manager.rs
leeyr338: Just for test, this would be `false`.

Makefile
leeyr338: ok

u2:
Before I have reviewed network-p2p, I have left some comments here https://github.com/u2/network-p2p/commits/master, I am not sure whether it's useful.

cita-network/src/node_manager.rs
```diff
@@ -0,0 +1,615 @@
...
+    pub fn handle(self, service: &mut NodesManager) {
+        trace!(
+            "[NodeManager] Send msg {:?} to {}, from key {}",
+            self.msg,
+            self.dst,
+            self.key
+        );
+        let msg_bytes: Vec<u8> = self.msg.try_into().unwrap();
+
+        let mut buf = BytesMut::with_capacity(4 + 4 + 1 + self.key.len() + msg_bytes.len());
```
leeyr338: from u2: It's better to use some constants.

cita-network/src/mq_agent.rs
```diff
@@ -0,0 +1,213 @@
...
+// useful, but WITHOUT ANY WARRANTY; without even the implied
+// warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+// PURPOSE. See the GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::network::{LocalMessage, NetworkClient};
+use crate::node_manager::{BroadcastReq, NodesManagerClient};
+use crossbeam_channel::{unbounded, Receiver, Sender};
```
yangby-cryptape: This will be a problem in future.

The channel send to [`fn start_pubsub()`] as input, should be same as the [definitions].
So, the changes in upstream (crate _`pubsub`_) will not effect downstream (code, here).

[`fn start_pubsub()`]: https://github.com/cryptape/cita-common/blob/8c25204/pubsub/src/lib.rs#L39-L78
[definitions]: https://github.com/cryptape/cita-common/blob/8c25204/pubsub/src/lib.rs#L26-L36

cita-network/src/synchronizer.rs
```diff
@@ -459,3 +452,74 @@ impl Synchronizer {
...
+    }
+
+    pub fn handle_remote_status(&self, msg: SynchronizerMessage) {
+        self.send_msg(msg);
+    }
+
+    pub fn handle_remote_response(&self, msg: SynchronizerMessage) {
+        self.send_msg(msg);
+    }
+
```
ouwenkg: These three functions work for the same thing?

cita-network/src/p2p_protocol/transfer.rs
```diff
@@ -0,0 +1,134 @@
...
+struct TransferProtocol {
+    proto_id: ProtocolId,
+    connected_session_ids: Vec<SessionId>,
+    network_client: NetworkClient,
+    nodes_mgr_client: NodesManagerClient,
+}
+
+impl ServiceProtocol for TransferProtocol {
+    fn init(&mut self, _control: &mut ServiceContext) {}
+
```
ouwenkg: This `init()` needed for test ?

driftluo:
There is a very important thing here that is not defined, probably because the user has ignored this thing because of the default implementation.

The selection process of the protocol opening is negotiated by the version number and name. These two things are also the key points for the compatibility of the protocol.

In the default implementation, the name is `/p2p/{proto_id}`, but the protocol name can theoretically be independent of the internal id. In this case, no matter what id is used in your internal implementation, you don’t need to reflect this during the negotiation process.

The version number determines whether the current client is compatible with the old version protocol. For example, A supports 0.0.1 B support [0.0.1, 0.2]. During the negotiation process, the 0.0.1 version of the protocol will be enabled.

cita-network/src/config.rs
```diff
@@ -37,32 +42,54 @@ impl NetConfig {
...
-        port = 40000
+        port = 4000
         enable_tls = true
+        max_connects = 4
+        id_card = 9
         [[peers]]
-        ip = "127.0.0.1"
-        port = 40001
-        common_name = "test1.cita"
+            ip = "0.0.0.0"
```
yangby-cryptape: It is weird. Why?

cita-network/src/main.rs
```diff
@@ -134,127 +111,62 @@ fn main() {
...
 
     // init app
-    // todo load config
     let matches = App::new("network")
         .version(get_build_info_str(true))
         .long_version(get_build_info_str(false))
         .author("Cryptape")
         .about("CITA Block Chain Node powered by Rust")
         .args_from_usage("-c, --config=[FILE] 'Sets a custom config file'")
+        .args_from_usage("-a, --address=[FILE] 'Sets an address file'")
```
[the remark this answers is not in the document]
leeyr338: I have a default value for now.

cita-network/src/synchronizer.rs
leeyr338: Yes, I want to distinguish between different semantics for caller.

cita-network/src/p2p_protocol/transfer.rs
leeyr338: This is a function needed by the trait `ServiceProtocol`, and I don't want to do something in it.

cita-network/src/mq_agent.rs
leeyr338: Yes, This should be the same define as `fn start_pubsub()`, I will use `pubsub::channel` instead.

cita-network/src/config.rs
leeyr338: Oh, you are saying the value of `ip` or the indentation of the code?

yangby-cryptape: I had a chat with @jerry-yu about it at lunch time, we both think the IP should not be `"0.0.0.0"`.

leeyr338: Got it!

kaikai1024:
Clear up your commits, please:

* fix clippy error for network
* fix fmt error
* rebase from develop
* code fix after rebase
* delete Cargo.toml.bak
* do not change makefile

Or we will use squash and merge.

Issue #243, opened 2019-02-19
Specify the path to of data and log.


kaikai1024:
## Description

Specify the path to of data and log.

## Steps to Reproduce

Omit

## Versions

develop

## Additional Information

----------------Chinese-------------

可指定 log 和 data 文件夹的路径

kaikai1024:
#221

leeyr338:
Has  someone work on this issue?

u2:
> Has someone work on this issue?

If you have more time, can you help with this task https://github.com/cryptape/cita-common/issues/177.

kaikai1024:
> Has someone work on this issue?

I will.

kaikai1024:
> > Has someone work on this issue?
> 
> If you have more time, can you help with this task [cryptape/cita-common#177](https://github.com/cryptape/cita-common/issues/177).

This's also me.

kaikai1024:
Later. 
new solution, check #283

PR #279, opened 2019-03-05
[WIP] Update image: cita run and cita create config.

kaikai1024:

Split cita_run image in env.sh:
* new cita run
* cita create config

#### Benefits
* Create cita config is independent.
* for #243 

#### Verification Process

Check two containers after create cita config and setup node.

#### Applicable Issues

#270

### DIFF
--- a/scripts/env.sh
+++ b/scripts/env.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-DOCKER_IMAGE="cita/cita-run:ubuntu-18.04-20181009"
+DOCKER_IMAGE="cita/cita-run:ubuntu-18.04-20190228"
 if [[ `uname` == 'Darwin' ]]
 then
     cp /etc/localtime $PWD/localtime
@@ -26,6 +26,11 @@ CONTAINER_NAME="cita_run${RELEASE_DIR//\//_}"
 WORKDIR=/opt/cita-run
 USER_ID=`id -u $USER`
 USER_NAME="user"
+
+if [[ "${1}" =~ "create_cita_config" ]]; then
+    CONTAINER_NAME="cita_config${RELEASE_DIR//\//_}"
+    DOCKER_IMAGE="cita/cita-run:config-ubuntu-18.04-20190228"
+fi
 
 if [ "${USER_ID}" = "0" ]; then
     USER_NAME="root"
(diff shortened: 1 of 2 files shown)


scripts/env.sh
```diff
@@ -27,6 +27,11 @@ WORKDIR=/opt/cita-run
 USER_ID=`id -u $USER`
 USER_NAME="user"
 
+if [[ "${1}" =~ "create_cita_config" ]]; then
```
yangby-cryptape: Why use `=~` ?

scripts/env.sh
```diff
@@ -27,6 +27,11 @@ WORKDIR=/opt/cita-run
...
+    CONTAINER_NAME="cita_config${RELEASE_DIR//\//_}"
```
yangby-cryptape: `"cita_config"` and `"cita_run"` should be a variable before define `CONTAINER_NAME`.
So, we can write `${RELEASE_DIR//\//_}` only once.

scripts/env.sh
```diff
@@ -27,6 +27,11 @@ WORKDIR=/opt/cita-run
...
+    DOCKER_IMAGE="cita/cita-run:config-ubuntu-18.04-20190228"
```
yangby-cryptape: May be `ubuntu-18.04-20190228` can be a variable.
```bash
ID="ubuntu-18.04-20190228"
NAME="cita/cita-run"
if [ ..omit .. ]; then
    VERSION="${ID}"
else
    VERSION="config-${ID}"
fi
DOCKER_IMAGE="${NAME}:${VERSION}"
```

kaikai1024:
> I did not check the build scripts for docker `cita-run`.
> 
> But I think `cita/cita-run:config-${id}` should base on `cita/cita-run:${id}`.
> They should have same layers, and `cita/cita-run:config-${id}` has one more layer than `cita/cita-run:${id}`.

That's my fault. 
It should be `cita:cita-run` & `cita: cita-config`, I'll fix that.

scripts/env.sh
kaikai1024: For the future that If we refactor the create_cita_config.

More detail:
I want to use the same `env.sh`, so i need to differentiate two image type.
if `./env.sh *create_cita_config* xx`, then will use cita-config image.

scripts/env.sh
kaikai1024: get

scripts/env.sh
kaikai1024: :clap:

kaikai1024:
Later. TBD
#283

PR #290, opened 2019-03-06
:art:Put all config toml in one folder

ouwenkg:
# Put all config toml in one folder

### Requirements
After this pr, the directory structure in node as following:
```
$ ls test-chain/0
configs logs data

$ ls test-chain/0/configs/
address    chain.toml      executor.toml      forever.toml  jsonrpc.toml  privkey
auth.toml  consensus.toml  forever_mock.toml  genesis.json  network.toml

$ ls test-chain/0/logs/
cita-auth.log  cita-bft.log  cita-chain.log  cita-executor.log  cita-forever.log  cita-jsonrpc.log  cita-network.log

$ ls test-chain/0/data/
authorities  nosql  statedb  txwal  wal
```
Some changes in submodule `cita-forever`

### DIFF
--- a/scripts/cita
+++ b/scripts/cita
@@ -15,6 +15,7 @@ NODE_NAME=$2
 NODE_PATH="$(pwd)/${NODE_NAME}"
 NODE_LOGS_DIR="${NODE_PATH}/logs"
 NODE_DATA_DIR="${NODE_PATH}/data"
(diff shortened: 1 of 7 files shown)


kaikai1024:
Compatibility: If restart the node after upgrading version, then what?

PR #296, opened 2019-03-07
remove unused db column

jerry-yu:
remove unused db column to save memory


kaikai1024:
Is that the db log?

btw, You can just @ citaman instead of one by one.

u2:
Is this compatible with the before version?

Issue #293, opened 2019-03-07
Temporarily remove the genesis code check


kaikai1024:
## Description

Temporarily  remove the genesis code check.

## Versions

develop

## Additional Information

-----------Chinese-------
暂时移除 genesis 检查中的 code 检查

kaikai1024:
#216

ouwenkg:
#283

PR #301, opened 2019-03-08
Remove genesis code check

ouwenkg:
# Remove genesis code check

### Requirements
Temporarily remove the genesis code check. 

#### Applicable Issues
#293

### DIFF
--- a/tests/compatibility/check_genesis.py
+++ b/tests/compatibility/check_genesis.py
@@ -20,11 +20,25 @@ def parse_arguments():
         required=True,
         help='The genesis file to be checked.')
 
+    check_code_parser = parser.add_mutually_exclusive_group(required=False)
+    check_code_parser.add_argument(
+        '--check_code',
+        dest='check_code',
+        action='store_true',
+        help='Check genesis contract code.')
+    check_code_parser.add_argument(
+        '--no_check_code',
+        dest='check_code',
+        action='store_false',
+        help='Do not check genesis contract code.'
+    )
+    parser.set_defaults(check_code=False)
+
     args = parser.parse_args()
     return args
 
 
-def check(old, new):
+def check(old, new, check_code):
     """ Check the new genesis is not changed. """
     old_alloc = old['alloc']
     new_alloc = new['alloc']
@@ -34,7 +48,7 @@ def check(old, new):
         if 'value' in old_alloc[addr]:
             continue
         # Check the code
-        if old_alloc[addr]['code'] != new_alloc[addr]['code']:
+        if check_code and old_alloc[addr]['code'] != new_alloc[addr]['code']:
             return False
         # Check the storage
         storage = old_alloc[addr]['storage']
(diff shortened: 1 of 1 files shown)


kaikai1024:
Add the info of applicable Issues.

yangby-cryptape:
A option is better.

ouwenkg:
> Provide an input parameter to control this feature is better.

Input parameters like cita version or what? Could you please make it more detailed?

kaikai1024:
like `skip audit` ?

yangby-cryptape:
For example:

- ```sh
  python check_genesis.py --check-code=false
  ```

- ```sh
  python check_genesis.py --check=storage,code
  ```

kaikai1024:
I see.

ouwenkg:
I design it again.

kaikai1024:
> I think there no need to check genesis code in this test. we have checked the storage in contract.

Why?

kaikai1024:
I'll do it.

yangby-cryptape:
I do not care about whether it is required or not.

I just do not like commented code in a [version control system].
- If this piece of code is redundant, we delete them.
- If we just have to remove it temporarily, we offer an option to control it.

Anyway, you both could keep going.

[version control system]: https://en.wikipedia.org/wiki/Revision_control

tests/compatibility/check_genesis.py
```diff
@@ -34,7 +38,7 @@ def check(old, new):
         if 'value' in old_alloc[addr]:
             continue
         # Check the code
-        if old_alloc[addr]['code'] != new_alloc[addr]['code']:
+        if check_code is True and old_alloc[addr]['code'] != new_alloc[addr]['code']:
```
kaikai1024: `if check_code and ..`

tests/compatibility/check_genesis.sh
```diff
@@ -14,5 +14,5 @@ cd ${BINARY_DIR} \
     --super_admin "0x0000000000000000000000000000000000000000" \
     --nodes "127.0.0.1:4000" \
 && python3 ${SOURCE_DIR}/tests/compatibility/check_genesis.py \
-    --genesis test-chain/0/genesis.json \
+    --genesis test-chain/0/genesis.json --check_code false \
```
kaikai1024: is that ok? are you try to check the value of `args.check_code`?

tests/compatibility/check_genesis.sh
```diff
@@ -14,5 +14,5 @@ cd ${BINARY_DIR} \
     --super_admin "0x0000000000000000000000000000000000000000" \
     --nodes "127.0.0.1:4000" \
 && python3 ${SOURCE_DIR}/tests/compatibility/check_genesis.py \
-    --genesis test-chain/0/genesis.json \
+    --genesis test-chain/0/genesis.json\
```
kaikai1024: white space

tests/compatibility/check_genesis.py
ouwenkg: That's a stupid mistake i made, i remembered it.

tests/compatibility/check_genesis.sh
ouwenkg: Done!

Issue #302, opened 2019-03-08
cita-executor crashes when receives staled `BlockWithProof`


keroro520:
## Description

cita-executor crashes when receives staled `BlockWithProof`.

[Here](https://gist.github.com/keroro520/063086929918ce3c3360fd5c3dc79b23) is a test-script to reproduce.

## Steps to Reproduce

1. Start CITA components manually. Let it grows a while.

    ```
    cd ${CHAIN_NAME}/0
    ${BINARY_DIR}/bin/cita-auth -c auth.toml &  
    ${BINARY_DIR}/bin/cita-bft -c consensus.toml -p privkey &  
    ${BINARY_DIR}/bin/cita-chain -c chain.toml &  
    ${BINARY_DIR}/bin/cita-executor -c executor.toml &  
    ${BINARY_DIR}/bin/cita-jsonrpc -c jsonrpc.toml &  
    ${BINARY_DIR}/bin/cita-network -c network.toml &  
    ```

2. Kill cita-executor and cita-chain. Then wait a minute.

3. Kill all rest CITA processes(cita-bft, cita-auth and so on).

4. `rm -rf ${NODE_PATH}/data/statedb`

5. Use `${BINARY_DIR}/bin/cita start ${NODE_PATH} trace` to restart CITA.

6. Expect cita-executor crashes over and over again.

**Expected behavior**: CITA behaviors OK after restarting.

**Actual behavior**: cita-executor crashed.

**Reproduce how often**: 80%

## Versions

Branch "develop"
commit "3c63cd3b9db8c4cb13e658d6eeeaddbf21713a04"

## Additional Information

keroro520:
cita-executor does not allow staled `BlockWithProof` at present.

But in reality, cita-bft would re-send staled `BlockWithProof` in some scenario.

So maybe we should make cita-executor tolerant this kind of faulty.

keroro520:
I have create a [branch](https://github.com/cryptape/cita/tree/fix/executor-tolerant-staled) to fix this issue, but fail to compile CITA at present.

So hope @ouwenkg help me to continue this bugfix ~

kaikai1024:
ok, thx

PR #306, opened 2019-03-11
Fix executor crashed

ouwenkg:
# Fix executor crashed

### Description
Fix panic usage in `is_proof_ok()` executor module.

#### Applicable Issues
#302


kaikai1024:
Cooperate on the `fix/executor-tolerant-staled`.

PR #314, opened 2019-03-13
Format py

ouwenkg:
# Format python code

Just format py. If necessary, add checker later.

#### Application issue
#315


kaikai1024:
Actually, you don't need to add independent in the image, you can check it on the stage and use another docker image with script installing the independent of lint.

kaikai1024:
cool

kaikai1024:
Create a new issue about ci. @ouwenkg

PR #321, opened 2019-03-13
Fix JSON-RPC CORS.

yangby-cryptape:
- The [`Access-Control-Allow-Origin`] is not right.
  Should be `Access-Control-Allow-Origin`, but we write `Origin`.

- The [`Access-Control-Allow-Origin`] should be `*`, `<origin>` or `null`.
  The empty string is not in the specification.

- Add a simple test to test if [`Access-Control-Allow-Origin`] is existed in JSON-RPC's responses.

[`Access-Control-Allow-Origin`]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#Syntax

### DIFF
--- a/cita-jsonrpc/src/http_header.rs
+++ b/cita-jsonrpc/src/http_header.rs
@@ -18,7 +18,7 @@
 use hyper::header::{HeaderMap, HeaderName, HeaderValue, InvalidHeaderValue};
 
 pub const ORIGIN_ANY_STR: &str = "*";
-pub const ORIGIN_NULL_STR: &str = "";
+pub const ORIGIN_NULL_STR: &str = "null";
 
 // values come from hyper 0.11
 pub const CONTENT_TYPE_PLAIN_TEXT_STR: &str = "text/plain; charset=utf-8";
@@ -174,7 +174,7 @@ mod test {
 
     #[test]
     fn test_origin_null() {
-        assert_eq!(Origin::null(), HeaderValue::from_static(""));
+        assert_eq!(Origin::null(), HeaderValue::from_static("null"));
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


tests/integrate_test/cita_jsonrpcTest.sh
```diff
@@ -34,6 +34,11 @@ invalid_data=-32600
 invalid_jsonrpc_method=-32601
 
 ./cita_start.sh
+
```
yangby-cryptape: This script is not in CI.
Need another issue to check these scripts and add them back to CI.

ouwenkg: :+1:  Let me have a try.

cita-jsonrpc/src/http_header.rs
```diff
@@ -192,9 +192,9 @@ mod test {
         vec![
             (Some("*"), HeaderValue::from_static("*")),
             (Some(" * "), HeaderValue::from_static("*")),
-            (None, HeaderValue::from_static("")),
-            (Some(""), HeaderValue::from_static("")),
-            (Some("   null "), HeaderValue::from_static("")),
```
u2: I do not understand this test case, could you help to explain for me?

yangby-cryptape: Test these:
https://github.com/cryptape/cita/blob/c8647954682bb68fd9707194692cb73c93ad2302/cita-jsonrpc/src/http_header.rs#L70-L76

u2: I see.

tests/integrate_test/cita_jsonrpcTest.sh
```diff
@@ -34,6 +34,11 @@ invalid_data=-32600
...
+# Check JSON-RPC CORS: Access-Control-Allow-Origin should be existed.
+has_cors=$(curl -i -X POST -d '{"jsonrpc":"2.0","method":"peerCount","params":[],"id":2}'  $IP:$PORT 2>/dev/null | grep -ic "^access-control-allow-origin: ")
```
yangby-cryptape: Suddenly, I realized, the test request should be:
```
curl -i -X OPTIONS $IP:$PORT 2>/dev/null | grep -ic "^access-control-allow-origin: "
```

All are OK, but this is better.

clearloop:
And except the docker's CORS, my browser still can't get response from the JSONRPC server, `cita-cli` and `curl` work well, I'm wondering happened.

The latest response is a `timeout` error and the truffle box, doesn't happy either....tried both `ws` and `http`.

btw, `cita-sdk` can work in the `node` environment.

Issue #326, opened 2019-03-14
cita-auth  crashes on performance test


leeyr338:
## Description
Run performance test `Counter_bench`  leads cita-auth run crashes, and get logs as below:
```
2019-03-14T02:20:57.678666211+08:00 - INFO - public block txs height 91 with 0 transactions
2019-03-14T02:21:00.677223636+08:00 - INFO - get block tx hashes for height 92
2019-03-14T02:21:00.677295443+08:00 - INFO - public block txs height 92 with 2 transactions
2019-03-14T02:21:04.063847069+08:00 - INFO - get block tx hashes for height 93
2019-03-14T02:21:04.063922839+08:00 - INFO - public block txs height 93 with 2 transactions
2019-03-14T02:21:06.674889646+08:00 - INFO - GetBlockTxn size: 2, origin: 1
2019-03-14T02:21:06.676471905+08:00 - INFO - Net >> BlockTxn: auth is not ready

// crashes here, and restarted by cita-forever.
2019-03-14T02:21:06.739075250+08:00 - INFO - CITA:auth
2019-03-14T02:21:06.739115113+08:00 - INFO - Version: v0.21rc-89-gc864795
2019-03-14T02:21:06.778715344+08:00 - INFO - Session initialized
2019-03-14T02:21:06.822011671+08:00 - INFO - Session initialized
2019-03-14T02:21:08.063911004+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.141921838+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.142061364+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.169812155+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.169926582+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.198717953+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.198803486+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.229030371+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.229158822+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.259385249+08:00 - WARN - Too frequent to send request!
```

**Expected behavior**: [What you expect to happen]
cita-auth do not crash.

**Actual behavior**: [What actually happens]
cita-auth crash.

**Reproduce how often**: 
100%

## Versions
v0.22.0-rc

## Additional Information
[...]

PR #328, opened 2019-03-14
Bugfix for auth crashes on performance test

leeyr338:
Bugfix for auth crashes on performance test: 
1. Do not return when auth not ready;
2. Do not broadcast all auth message.

#326

### DIFF
--- a/cita-auth/src/handler.rs
+++ b/cita-auth/src/handler.rs
@@ -377,7 +377,6 @@ impl MsgHandler {
         // we will send req for all height
         // so don't too frequent
         if check && self.history_heights.is_too_frequent() {
-            warn!("Too frequent to send request!");
             return;
         }
         trace!(
@@ -467,7 +466,6 @@ impl MsgHandler {
                     routing_key!(Consensus >> VerifyBlockReq) => {
                         if !self.is_ready() {
                             info!("Net/Consensus >> CompactProposal: auth is not ready");
-                            return;
                         } else {
                             self.deal_signed_proposal(msg);
                         }
@@ -475,7 +473,6 @@ impl MsgHandler {
                     routing_key!(Net >> BlockTxn) => {
                         if !self.is_ready() || self.verify_block_req.is_none() {
                             info!("Net >> BlockTxn: auth is not ready");
-                            return;
                         } else {
                             let verify_block_req = self.verify_block_req.clone();
                             if let Some(verify_block_req) = verify_block_req {
--- a/cita-network/src/network.rs
+++ b/cita-network/src/network.rs
@@ -16,12 +16,12 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use crate::mq_agent::{MqAgentClient, PubMessage};
-use crate::node_manager::{BroadcastReq, GetPeerCountReq, NodesManagerClient};
+use crate::node_manager::{BroadcastReq, GetPeerCountReq, NodesManagerClient, SingleTxReq};
 use crate::synchronizer::{SynchronizerClient, SynchronizerMessage};
 use libproto::router::{MsgType, RoutingKey, SubModules};
 use libproto::routing_key;
 use libproto::snapshot::{Cmd, Resp, SnapshotResp};
-use libproto::{Message as ProtoMessage, Response};
+use libproto::{Message as ProtoMessage, OperateType, Response};
 use libproto::{TryFrom, TryInto};
 use logger::{error, info, trace, warn};
 use pubsub::channel::{unbounded, Receiver, Sender};
@@ -134,10 +134,11 @@ impl LocalMessage {
             }
             routing_key!(Chain >> SyncResponse) => {
                 let msg = ProtoMessage::try_from(&self.data).unwrap();
-                service.nodes_mgr_client.broadcast(BroadcastReq::new(
+                send_message(
+                    &service.nodes_mgr_client,
                     routing_key!(Synchronizer >> SyncResponse).into(),
                     msg,
-                ));
+                );
             }
             routing_key!(Jsonrpc >> RequestNet) => {
                 self.reply_rpc(&self.data, service);
@@ -288,6 +289,24 @@ impl RemoteMessage {
             _ => {
                 error!("[Network] Unexpected key {} from Remote", self.key);
             }
+        }
+    }
+}
+
+pub fn send_message(nodes_mgr_client: &NodesManagerClient, key: String, msg: ProtoMessage) {
+    let operate = msg.get_operate();
+
+    match operate {
+        OperateType::Broadcast => {
+            nodes_mgr_client.broadcast(BroadcastReq::new(key, msg));
+        }
+        OperateType::Single => {
+            let dst = msg.get_origin();
+            nodes_mgr_client.send_message(SingleTxReq::new(dst as usize, key, msg));
+        }
+        OperateType::Subtract => {
+            // FIXME: Support subtract broadcast if necessary.
+            warn!("[MqAgent] Subtract broadcast does not support yet!");
         }
     }
 }
--- a/cita-network/src/network.rs
+++ b/cita-network/src/network.rs
@@ -305,8 +305,9 @@ pub fn send_message(nodes_mgr_client: &NodesManagerClient, key: String, msg: Pro
             nodes_mgr_client.send_message(SingleTxReq::new(dst as usize, key, msg));
         }
         OperateType::Subtract => {
-            // FIXME: Support subtract broadcast if necessary.
-            warn!("[MqAgent] Subtract broadcast does not support yet!");
+            // FIXME: Support subtract broadcast if necessary, just use broadcast instead.
+            warn!("[MqAgent] Subtract broadcast does not support yet, use broadcast instead!");
+            nodes_mgr_client.broadcast(BroadcastReq::new(key, msg));
         }
     }
 }
--- a/cita-network/src/cita_protocol.rs
+++ b/cita-network/src/cita_protocol.rs
@@ -52,7 +52,7 @@ const NETMSG_START: u64 = 0xDEAD_BEEF_0000_0000;
 // And this will consume "4 + 4 + 1" fixed-lengths of the frame.
 pub const CITA_FRAME_HEADER_LEN: usize = 4 + 4 + 1;
 
-fn opt_bytes_extend(buf: &mut BytesMut, data: &[u8]) {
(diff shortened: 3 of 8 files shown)


cita-auth/src/handler.rs
```diff
@@ -377,7 +377,6 @@ impl MsgHandler {
         // we will send req for all height
         // so don't too frequent
         if check && self.history_heights.is_too_frequent() {
-            warn!("Too frequent to send request!");
```
kaikai1024: Why delete it?
info or debug or trace?

kaikai1024:
Why this can fix the issue? A little confused.

cita-auth/src/handler.rs
leeyr338: This log is meaningless. Even if the program enters this logic, it does not mean `Too frequent to send request.`

leeyr338:
> Why this can fix the issue? A little confused.

First, the log show us that auth exit main loop after `INFO - Net >> BlockTxn: auth is not ready`.
```
2019-03-14T02:21:06.674889646+08:00 - INFO - GetBlockTxn size: 2, origin: 1
2019-03-14T02:21:06.676471905+08:00 - INFO - Net >> BlockTxn: auth is not ready

// crashes here, and restarted by cita-forever.
2019-03-14T02:21:06.739075250+08:00 - INFO - CITA:auth
```
That is because we have a `return` in the logic:
```rust
if !self.is_ready() || self.verify_block_req.is_none() {
                            info!("Net >> BlockTxn: auth is not ready");
                            return;
                        } 
```
After being several times restarted by `forever`, the `Auth` service will be lost in `node0`, this causes the front-end of the test-case goes into wrong logic.

So, remove `return` logic in `Auth` can fix this situation.

Secondly, I notice that `Auth` received some  `BlockTxs` from a wrong `destination node`. That because I broadcast the request to all nodes, but actually it only needs to be sent to a exactly consensus node. This will cause a lot of extra packet and slow down the network.

So, handling these request packets correctly is very important.

cita-network/src/network.rs
```diff
@@ -291,3 +292,21 @@ impl RemoteMessage {
...
+
+    match operate {
+        OperateType::Broadcast => {
+            nodes_mgr_client.broadcast(BroadcastReq::new(key, msg));
+        }
+        OperateType::Single => {
+            let dst = msg.get_origin();
+            nodes_mgr_client.send_message(SingleTxReq::new(dst as usize, key, msg));
+        }
+        OperateType::Subtract => {
```
yangby-cryptape: At least, broadcast to all.

cita-auth/src/handler.rs
u2: If `is_too_frequent` is useless, we should create another PR to fix it.

u2:
> > Why this can fix the issue? A little confused.
> 
> First, the log show us that auth exit main loop after `INFO - Net >> BlockTxn: auth is not ready`.
> 
> ```
> 2019-03-14T02:21:06.674889646+08:00 - INFO - GetBlockTxn size: 2, origin: 1
> 2019-03-14T02:21:06.676471905+08:00 - INFO - Net >> BlockTxn: auth is not ready
> 
> // crashes here, and restarted by cita-forever.
> 2019-03-14T02:21:06.739075250+08:00 - INFO - CITA:auth
> ```
> That is because we have a `return` in the logic:
> 
> ```rust
> if !self.is_ready() || self.verify_block_req.is_none() {
>                             info!("Net >> BlockTxn: auth is not ready");
>                             return;
>                         } 
> ```
> 
> After being several times restarted by `forever`, the `Auth` service will be lost in `node0`, this causes the front-end of the test-case goes into wrong logic.
> 
> So, remove `return` logic in `Auth` can fix this situation.
> 
> Secondly, I notice that `Auth` received some `BlockTxs` from a wrong `destination node`. That because I broadcast the request to all nodes, but actually it only needs to be sent to a exactly consensus node. This will cause a lot of extra packet and slow down the network.
> 
> So, handling these request packets correctly is very important.

Yes, it should not be `return`, should be `continue`.

cita-network/src/network.rs
leeyr338: That sounds more reasonable.

cita-auth/src/handler.rs
leeyr338: But this also great influence on this PR. Because this log print too frequent, they will consume a lot of cpu time, resulting in slower auth execution.
```
2019-03-14T02:21:06.822011671+08:00 - INFO - Session initialized
2019-03-14T02:21:08.063911004+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.141921838+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.142061364+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.169812155+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.169926582+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.198717953+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.198803486+08:00 - WARN - Too frequent to send request!
2019-03-14T02:21:08.229030371+08:00 - WARN - Too frequent to send request!
```
So I think is better to handle this issue in this PR.

driftluo:
https://github.com/cryptape/cita/blob/develop/cita-network/src/cita_protocol.rs#L71:L88 useless code should delete

cita-network/src/network.rs
kaikai1024: What's subtract broadcast?

driftluo:
https://github.com/cryptape/cita/blob/a00ec653934f0bb509e2f2431457464be74cd676/cita-network/src/node_manager.rs#L590-L596

issue here: https://github.com/carllerche/bytes/issues/86

Bytes `to_vec()` will copy all data, if the amount of data is large, it will be very cumbersome.

we can change `pubsub_message_to_network_message` to this:
```rust
pub fn pubsub_message_to_network_message(buf: &mut Vec<u8>, msg: Option<(String, Vec<u8>)>)
```

and `opt_bytes_extend` function to this, I think unsafe in this function can be removed, should test it:
```rust
fn opt_bytes_extend(buf: &mut Vec<u8>, data: &[u8]) {
    buf.reserve(data.len());
    buf[..data.len()].copy_from_slice(data);
}
```

cita-network/src/network.rs
yangby-cryptape: > What's subtract broadcast?

- `Broadcast()` means to all.
- `Single()` means to the specify node.
- `Subtract()` means to all except the specify node.

leeyr338: Broadcast to all nodes except the specified one.

leeyr338:
> [cita/cita-network/src/node_manager.rs](https://github.com/cryptape/cita/blob/a00ec653934f0bb509e2f2431457464be74cd676/cita-network/src/node_manager.rs#L590-L596)
> 
> Lines 590 to 596 in [a00ec65](/cryptape/cita/commit/a00ec653934f0bb509e2f2431457464be74cd676)
> 
>   
>  let mut buf = 
>      BytesMut::with_capacity(CITA_FRAME_HEADER_LEN + self.key.len() + msg_bytes.len()); 
>  pubsub_message_to_network_message(&mut buf, Some((self.key, msg_bytes))); 
>  if let Some(ref mut ctrl) = service.service_ctrl { 
>      // FIXME: handle the error! 
>      let _ = ctrl.send_message(self.dst, 1, buf.to_vec()); 
> issue here: [carllerche/bytes#86](https://github.com/carllerche/bytes/issues/86)
> 
> Bytes `to_vec()` will copy all data, if the amount of data is large, it will be very cumbersome.
> 
> we can change `pubsub_message_to_network_message` to this:
> 
> ```rust
> pub fn pubsub_message_to_network_message(buf: &mut Vec<u8>, msg: Option<(String, Vec<u8>)>)
> ```
> 
> and change `opt_bytes_extend` function to this, I think unsafe in this function can be removed, should test it:
> 
> ```rust
> fn opt_bytes_extend(buf: &mut Vec<u8>, data: &[u8]) {
>     buf.reserve(data.len());
>     buf[..data.len()].copy_from_slice(data);
> }
> ```

`unsafe` code can get higher performance. So we just use this code.

cita-network/src/network.rs
kaikai1024: Got it.

kaikai1024:
> unsafe code can get higher performance. So we just use this code.

That's not a acceptable reason for me.  i believe `safety > performance`, we need higher performance on the base of safety. 
And, I though we use `unsafe` when we **have to** use it.

leeyr338:
> > unsafe code can get higher performance. So we just use this code.
> 
> That's not an acceptable reason for me. i believe `safety > performance`, we need higher performance on the base of safety.
> And, I though we use `unsafe` when we **have to** use it.

I have talked to driftluo, it is safe enough to use the `unsafe` keyword for this logic, but can get a higher performance.

driftluo:
I looked at this function, if modified to the safe function `extend_from_slice`, the performance will be reduced from O(1) to O(n) because it is written one by one, [reference](https://doc.rust-lang.org/src/alloc/vec.rs.html#1843-1861). 

Here `unsafe` use should be no problem, just a bit ugly.

driftluo:
If we really care about this, we can use the [append](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) function, the principle is the same as the unsafe code we are using now, but export Safe interface

u2:
Is the unsafe code related to this PR? It's another problem.

leeyr338:
> I looked at this function, if modified to the safe function `extend_from_slice`, the performance will be reduced from O(1) to O(n) because it is written one by one, [reference](https://doc.rust-lang.org/src/alloc/vec.rs.html#1843-1861).
> 
> Here `unsafe` use should be no problem, just a bit ugly.

Thanks @driftluo . This logic is the key-path for every messages transferring to other nodes, so this performance improvement here is also important.

leeyr338:
> Is the unsafe code related to this PR? It's another problem.

No, just a review comment from @driftluo .

u2:
> If we really care about this, we can use the [append](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.append) function, the principle is the same as the unsafe code we are using now, but export Safe interface

```
    unsafe {
        buf.bytes_mut()[..data.len()].copy_from_slice(data);
        buf.advance_mut(data.len());
    }
```

The intention of this code is `append`. I have got it before, it's hard to read.

kaikai1024:
Also pre-v0.22.0 branch @leeyr338

PR #336, opened 2019-03-18
Add get release version from RPC

luqz:
# Add get  release version from RPC

### Requirements
rely  cryptape/cita-common#226

#### Verification Process
1.  Add enable_version = true Or enable_version = false Or Nothing;
2.  POST getVersion to RPC
```
curl -X POST --data '{"jsonrpc":"2.0","method":"getVersion","params":[],"id":3}'  127.0.0.1:1337
```
#### Applicable Issues
#282 Get release version info of all modules from RPC.


Kayryu:
> We just get `JSON-RPC`'s binary version, so I don't think we need so mush logic, just like:
> 
> 1. Using a `RountingKey` to send the message from one thread to another;
> 2. A new thread to handle only the `getVersion` logic.
> 
> Is it possible to send the version back just after we get the command from `http`?

the origin RPC logic:
user ---> (1)http/ ws  ----> (2) dispach -----> (3)mq ----> (4) submodules ---->mq----> http/ws
(1): need change jsonrpc-types,

now logic:
user ---> (1)http/ ws  ----> (2) dispach -----> (3) self_handler ----> http/ws.
we hook request at (2) step.  and dispach requst to self responser.

cita-jsonrpc/Cargo.toml
```diff
@@ -18,9 +18,9 @@ util = { git = "https://github.com/cryptape/cita-common.git", branch = "develop"
 error = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
 logger = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
 pubsub = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-libproto = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-jsonrpc-types = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-jsonrpc-proto = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
+libproto = { git = "https://github.com/luqz/cita-common.git", branch = "develop" }
```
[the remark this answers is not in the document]
luqz: I will delete the second commit after the pr of cita-common was accept.

luqz:
> We just get `JSON-RPC`'s binary version, so I don't think we need so mush logic, just like:
> 
> 1. Using a `RountingKey` to send the message from one thread to another;
> 2. A new thread to handle only the `getVersion` logic.
> 
> Is it possible to send the version back just after we get the command from `http`?

1. I can't find a better place to deal with both http and websocket request.
2. I think the logic would be more complex if I handle the request of each type.

leeyr338:
> > We just get `JSON-RPC`'s binary version, so I don't think we need so mush logic, just like:
> > 
> > 1. Using a `RountingKey` to send the message from one thread to another;
> > 2. A new thread to handle only the `getVersion` logic.
> > 
> > Is it possible to send the version back just after we get the command from `http`?
> 
> 1. I can't find a better place to deal with both http and websocket request.
> 2. I think the logic would be more complex if I handle the request of each type.

@zeroqn , would you please review this PR for us.

rainchen:
This is a new feature, should add a UT or IT test for this feature.

zeroqn:
If jsonrpc's version is enough, then we don't need to go around mq.
There is ```get_build_info_str``` to get jsonrpc's version. 
IMO, handle ```(&Method::GET, "/version")``` and directly return that version should be enough.

luqz:
> If jsonrpc's version is enough, then we don't need to go around mq.
> There is `get_build_info_str` to get jsonrpc's version.
> IMO, add `(&Method::GET, "/version")` and directly return that version should be enough.

in this way, how to deal websocket call?

zeroqn:
> > If jsonrpc's version is enough, then we don't need to go around mq.
> > There is `get_build_info_str` to get jsonrpc's version.
> > IMO, add `(&Method::GET, "/version")` and directly return that version should be enough.
> 
> in this way, how to deal websocket call?

There is a ```Sender```, you can directly send json string back.

Kayryu:
> If jsonrpc's version is enough, then we don't need to go around mq.
> There is `get_build_info_str` to get jsonrpc's version.
> IMO, add `(&Method::GET, "/version")` and directly return that version should be enough.

1.  If return directly by http without mq, we should consider about batch request. but it looks hard to implement.
2. Like [Ethereum](https://github.com/ethereum/wiki/wiki/JSON-RPC#curl-examples-explained) use `Method::POST` almost,  `(&Method::GET, "/version")`  isn't fit to [JSONRPC specification](https://www.jsonrpc.org/specification#id2), we should pass with method, id, params.  
3. we should implement same logic in protocal of http and websock.

i agrese with your idea，so what's better?

zeroqn:
Yea, batch request could be tricky.

cita-jsonrpc/src/helper.rs
```diff
@@ -40,6 +40,7 @@ pub fn select_topic(method: &str) -> String {
     match method {
         "peerCount" => routing_key!(Jsonrpc >> RequestNet).into(),
         "sendRawTransaction" | "sendTransaction" => routing_key!(Jsonrpc >> RequestNewTx).into(),
+        "getVersion" => routing_key!(Jsonrpc >> RequestRpc).into(),
```
ouwenkg: I have some confusions, could you explain for me?
Could we `getVersion` though branch `_ => routing_key!(Jsonrpc >> Request).into(),` rather than new logic branch? 

Thinking about how we query `getBlockNumber`, `getMetaData`

cita-jsonrpc/src/responser.rs
```diff
@@ -0,0 +1,83 @@
...
+    config: Config,
+}
+
+impl Responser {
+    pub fn new(config: Config, tx: Sender<(String, Vec<u8>)>) -> Responser {
+        Responser { tx, config }
+    }
+
+    pub fn reply_request(&self, _: RoutingKey, mut msg: Message) {
+        let req_opt = msg.take_request();
```
ouwenkg: `reply_request(&self, _: RoutingKey, mut msg: Message)`
According to my view, If the second parameter is useless, we could del it. if we need it actually,  a variable name should be given.

cita-jsonrpc/src/helper.rs
luqz: Requests will be dispatched to chain module  with routing_key!(Jsonrpc >> Request), it will be more complicated to sort out if we want to deal with it in jsonrpc module.

cita-jsonrpc/src/responser.rs
luqz: Yes, it's useless in this case. Considering scalability, I think we should keep it.

kaikai1024:
Please rebase the latest develop branch.
And sorry for the ci problem, Still communicate with circel.

There is a solution:

I give you a contribute permission(that means you can push your branch into cryptape/cita),What do you think? or we can wait a few days. @luqz

kaikai1024:
You can push the branch to the cryptape/cita now, then you can ref this at a new pr

PR #351, opened 2019-03-21
Add getVersion

luqz:
#336 Add get release version from RPC


luqz:
> * The comment should be like this: `// The comments`
> * Would you please add a test case at `tests/interfaces/rpc` folder?

@Kayryu will do it later.

cita-jsonrpc/src/main.rs
```diff
@@ -98,6 +98,7 @@ mod http_server;
 mod mq_handler;
 mod mq_publisher;
 mod response;
+mod responser;
```
zeroqn: ```responser``` is ambiguity, hard to distinguish it from ```response```.
Need a better name, soliloquy?

cita-jsonrpc/src/main.rs
```diff
@@ -189,14 +195,19 @@ fn main() {
...
                 let (topic, req): (String, reqlib::Request) = res;
-                forward_service(
-                    topic,
-                    req,
-                    &mut new_tx_request_buffer,
-                    &mut time_stamp,
-                    &tx_pub,
-                    &tx_flow_config,
-                );
+                if RoutingKey::from(&topic) == routing_key!(Jsonrpc >> RequestRpc) {
```
zeroqn:
```suggestion
use ```match```
```

cita-jsonrpc/src/responser.rs
```diff
@@ -0,0 +1,83 @@
...
+use jsonrpc_types::rpctypes::SoftwareVersion;
+use jsonrpc_types::ErrorCode;
+use libproto::protos::response::Response;
+use libproto::router::{MsgType, RoutingKey, SubModules};
+use libproto::Message;
+use libproto::TryInto;
+use pubsub::channel::Sender;
+use serde_json;
+
+pub struct Responser {
```
zeroqn: maybe better
```
pub struct Soliloquy {
    config: Config,
}

impl Soliloquy {
    pub fn new(config: Config) -> Self {
        Soliloquy { config }
    }

    pub fn handle(&self, msg_bytes: &[u8]) -> Message {
        let maybe_msg = Message::try_from(msg_bytes);
        let mut response = Response::new();

        if let Ok(Some(req)) = maybe_msg.map(|mut msg| msg.take_request()) {
            response.set_request_id(req.request_id);
            let enabled_version = self.config.enable_version.unwrap_or(false);
            debug!("getVersion enabled:{}", enabled_version);

            match req.req {
                Some(software_version(_)) if enabled_version => {
                    let version = get_build_info_str(true);
                    let vec: Vec<&str> = version.split('-').collect();
                    let version = vec[0].to_string();
                    if let Ok(json_ver) = serde_json::to_value(SoftwareVersion::new(version)) {
                        response.set_software_version(json_ver.to_string());
                    } else {
                        response.set_code(ErrorCode::InternalError.code());
                        response.set_error_msg(ErrorCode::InternalError.description());
                    }
                }
                _ => {
                    response.set_code(ErrorCode::MethodNotFound.code());
                    response.set_error_msg(ErrorCode::MethodNotFound.description());
                }
            }
        } else {
            warn!("receive unexpected data");
            response.set_code(ErrorCode::InvalidRequest.code());
            response.set_error_msg(ErrorCode::InvalidRequest.description());
        }

        response.into()
    }
}
```

cita-jsonrpc/src/main.rs
```diff
@@ -206,6 +217,15 @@ fn main() {
         }
     });
 
+    // response RequestRpc
+    let responser = Responser::new(config.clone(), tx_from_mq);
```
zeroqn: change a bit
```
    let soli_config = config.clone();
    thread::spawn(move || {
        let soliloquy = Soliloquy::new(soli_config);

        loop {
            if let Ok((_, msg_bytes)) = rx.recv() {
                let resp_msg = soliloquy.handle(&msg_bytes);
                let _ = soli_resp_tx.send((
                    routing_key!(Jsonrpc >> Response).into(),
                    resp_msg.try_into().unwrap(),
                ));
            }
        }
    });
```

Issue #354, opened 2019-03-22
Both nodes run normal when copy from a node without changing address and privkey ## 手动复制一个节点，不修改address和privkey,目前这两个节点都正常出块


mingxiaowu:
## Steps to Reproduce

1. stop节点 0，复制节点 0 为 4，启动节点 0；
2. 节点 4 的：删除 data 和 log；修改 network.toml, executor.toml, jsonrpc.toml, .env；不修改 address 和privkey；
3. 启动节点 4；

**Expected behavior**:
 
节点 4 和节点 0 的 address/privkey 相同，所以节点 4 应该被 kill

## Versions

v0.22.0-rc

leeyr338:
一个区块链网络没有权限去`kill`一个节点，网络能做的仅是**阻止它加入这个网络**。

kaikai1024:
Bft will treat them as a same node, right? @KaoImin

KaoImin:
> Bft will treat them as a same node, right? @KaoImin

Yes

leeyr338:
We have to fix this bug in `network`, make `node 4` cannot access into the block-chain. I am working on this.

PR #356, opened 2019-03-24
bugfix: refuse duplicate nodes

leeyr338:
Bugfix for #354.

### DIFF
--- a/cita-network/src/node_manager.rs
+++ b/cita-network/src/node_manager.rs
@@ -20,15 +20,13 @@ use crate::config::NetConfig;
 use crate::p2p_protocol::transfer::TRANSFER_PROTOCOL_ID;
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use cita_types::Address;
-use discovery::RawAddr;
 use fnv::FnvHashMap;
 use libproto::{Message as ProtoMessage, TryInto};
 use logger::{debug, error, info, trace, warn};
 use pubsub::channel::{select, tick, unbounded, Receiver, Sender};
-
+use rand;
 use std::{
     collections::HashMap,
-    collections::HashSet,
     convert::Into,
     io::Cursor,
     net::{SocketAddr, ToSocketAddrs},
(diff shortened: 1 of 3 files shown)


cita-network/src/node_manager.rs
```diff
@@ -41,13 +39,28 @@ use tentacle::{
 pub const DEFAULT_MAX_CONNECTS: usize = 666;
 pub const DEFAULT_PORT: usize = 4000;
 pub const CHECK_CONNECTED_NODES: Duration = Duration::from_secs(3);
+
+// Score uses to manage known_nodes list. If a node has too low score, do not dial it again.
```
kaikai1024: Is the score system a part of p2p or we add it just to `refuse-duplicate-nodes`?

cita-network/src/node_manager.rs
```diff
@@ -188,18 +244,47 @@ impl Default for NodesManager {
...
-            repeated_connections: HashSet::default(),
         }
     }
 }
 
+#[derive(Debug)]
+pub struct NodeStatus {
+    // score: Score for a node, it will affect whether the node will be chosen to dail again,
+    // or be deleted from the known_addresses list. But for now, it useless.
+    pub score: i32,
```
kaikai1024: score is useless, why we use it?

kaikai1024:
Would you please point out the codes that refusing the duplicate nodes?

cita-network/src/node_manager.rs
leeyr338: We have to use it to fix the problem.
The problem of `refused duplicate nodes` is a little complicated:
- First, need to find a node is a `duplicate node`. I design a `init message` to exchange their address to find it.
- Second, when the server found a `duplicate node`, it will disconnected the `duplicate node`, but
- Third, the `duplicate node` will dial all the time after it was disconnected by server, so I design `score` to handle this.

kaikai1024: >  But for now, it useless.
So `Third` is in the todo list, right?

leeyr338: Not exactly. Maybe I missed some information, the score design not just use for detecting `duplicated node`, but also `duplicated session (connection)`.
- A `duplicated session (connection)` is session connected to the same node.  For example, there are two nodes A and B, A dial to B will open a session; And B dial to A, It will open a `duplicated session` in this situation. 
- A `duplicated node` means two nodes contains the same `address` and `private key`.

driftluo:
Discovery, peer store, scoring mechanism, these things are almost indispensable functions on the p2p network.

This pr is problematic to the processing of `knows_addr`, should not use `remove` to delete, scoring and store are the means of `knows_addr`. Subsequent need to implement the peer store mechanism. #355

leeyr338:
> Would you please point out the codes that refusing the duplicate nodes?

After receiving the `init message` from `transfer` protocol, it will send a `AddConnectedNodeReq` message to `NodesManager`, means that this node is preparing to add a connected node. see the code `transfer.rs:L131`:
```rust
if key.eq(&"network.init".to_string()) {
                let msg = InitMsg::from(message);
                let req = AddConnectedNodeReq::new(session.id, session.ty, msg);
                self.nodes_mgr_client.add_connected_node(req);
                return;
            }
```
And then, you can get into the code of `handle` in `AddConnectedNodeReq`, here will handle how to add a `connected node`. see `node_manger.rs:L435`:
```rust
pub fn handle(self, service: &mut NodesManager) {
        if let Some(repeated_id) = service.connected_peer_keys.get(&self.init_msg.peer_key) {
            // Repeated connected, it can a duplicated connected to the same node, or a duplicated
            // node connected to this server. But in either case, disconnect this session.
            // In P2P encrypted communication mode, the repeated connection will be detected by
            // P2P framework, handling this situation by sending a `AddRepeatedNodeReq` message to
            // NodesManager. See the `handle` in `AddRepeatedNodeReq` for more detail.
            info!(
                "[NodeManager] New session [{:?}] repeated with [{:?}], disconnect this session.",
                self.session_id, *repeated_id
            );

            if let Some(ref mut ctrl) = service.service_ctrl {
                let _ = ctrl.disconnect(self.session_id);
            }
        }
...
}
```
And the first logic is handle `Repeated connected` which I have talked about, this can be a `duplicated node` or `duplicated connected`. We just `disconnected` the session.

After `session disconnected`, we can receive a `SessionClose` event from P2P framework, the will lead the code go into the `handle` of `DelConnectedNodeReq`.

leeyr338:
> Discovery, peer store, scoring mechanism, these things are almost indispensable functions on the p2p network.
> 
> This pr is problematic to the processing of `knows_addr`, should not use `remove` to delete, scoring and store are the means of `knows_addr`. Subsequent need to implement the peer store mechanism. #355

Yes, delete the `known_addrs` will get the bug describe in #355. I will fix this bug just after this PR.

PR #363, opened 2019-03-26
Bugfix for node lost

leeyr338:
Bugfix for node lost: #355 .
Fix this bug needs several steps:
- DO NOT remove node from `known_addrs` when dialed error;
- Use `score` mechanism:
 -> Deducts 10 score for each `Dial`;
 -> Deducts 25 score for each `Dial Error`;
 -> Deducts 20 score for each `Disconnected by server`;
 -> Add 5 score for every `dialing round` if the node keep on line; so If a node keep on line, it will get `FULL_SCORE` very fast.
-> Gives a `Time sugar score (2 : nodes was configured in config file, and 1 : nodes was discovered by P2P framework )` when a node's score less than `MIN_DIALING_SCORE`;

### DIFF
--- a/cita-network/src/node_manager.rs
+++ b/cita-network/src/node_manager.rs
@@ -27,6 +27,7 @@ use pubsub::channel::{select, tick, unbounded, Receiver, Sender};
 use rand;
 use std::{
     collections::HashMap,
+    collections::HashSet,
     convert::Into,
     io::Cursor,
     net::{SocketAddr, ToSocketAddrs},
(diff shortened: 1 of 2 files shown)


leeyr338:
> > -> Deducts 10 score for each Dial;
> > -> Deducts 25 score for each Dial Error;
> > -> Deducts 20 score for each Disconnected by server;
> > -> Add 5 score for every dialing round if the node keep on line; so If a node keep on line, it will get FULL_SCORE very fast.
> > -> Gives a Time sugar score (2 : nodes was configured in config file, and 1 : nodes was discovered by P2P framework ) when a node's score less than MIN_DIALING_SCORE;
> 
> That's should be a comment of the code.

OK!

kaikai1024:
> OK!

I already noticed that

cita-network/src/node_manager.rs
```diff
@@ -794,3 +876,18 @@ impl GetPeerCountReq {
...
+    addr: SocketAddr,
+}
+
+impl ConnectedSelfReq {
+    pub fn new(addr: SocketAddr) -> Self {
+        ConnectedSelfReq { addr }
+    }
+
+    pub fn handle(self, service: &mut NodesManager) {
+        service.self_addr = Some(self.addr);
```
driftluo: If you listen to multiple addresses or are mapped to multiple addresses, the processing here is problematic.

Can be modified later, CITA do not currently support multi-address listening

leeyr338: OK

PR #367, opened 2019-03-26
support config to getVersion and stdout

luqz:


cita-jsonrpc/src/soliloquy.rs
```diff
@@ -68,3 +68,140 @@ impl Soliloquy {
...
+
+        "#;
+
+        let response = get_response(toml_str.to_string());
+        assert_eq!(
+            response.get_software_version().contains("softwareVersion"),
+            true
+        );
+    }
+}
```
ouwenkg: Maybe two separated tests is better?

scripts/cita
```diff
@@ -162,11 +162,11 @@ do_start() {
 
     # Start cita-forever
     if [ -z ${debug} ]; then
-        cita-forever -c ${config} start > /dev/null 2>&1
+        cita-forever -c ${config} start
```
ouwenkg: why do this changes?

luqz: It wouldn't be logged to console if we don't change this.

kaikai1024: Maybe just remove `> /dev/null`.
`2>&1` is that error message out to the console.

cita-jsonrpc/src/soliloquy.rs
luqz: OK.

scripts/cita
luqz: OK.

scripts/create_cita_config.py
```diff
@@ -191,13 +191,24 @@ def template_create_from_arguments(self, args, contracts_dir_src,
...
+
         jsonrpc_config = os.path.join(self.configs_dir, 'jsonrpc.toml')
         with open(jsonrpc_config, 'rt') as stream:
             jsonrpc_data = toml.load(stream)
             jsonrpc_data['http_config']['listen_port'] \
                 = str(args.jsonrpc_port)
             jsonrpc_data['ws_config']['listen_port'] \
                 = str(args.ws_port)
+            if args.enable_version:
+                jsonrpc_data['enable_version'] = True
```
ouwenkg: I am confused the default `enable_version` is false or true?

Several steps this config works:

1. When cita compiled, this config located in  `target/install/scripts/config_tool/config_example`

2. When nodes created, the first step is copying config in scripts dir to template dir, 

3. When nodes created, the second step is rewriting config in template dir to node dir.

In your implecation, you set `enable_version` false ` in step 1, then set it true in step 2.

luqz: The default enable_version is false.
In step 2, only if --enable_version is present when we run create_cita_config.py, enable_version in jsonrpc.toml would be set to true.

ouwenkg: ok

Issue #359, opened 2019-03-25
The test of getVersion.


kaikai1024:
## Description

Add the unit test of getVersion

## Additional Information

#351

leeyr338:
Can anyone help to support this JSON-RPC interface for cita-cli?
https://github.com/cryptape/cita-cli/issues/106

kaikai1024:
Use `curl` to test if if works.

I mean the `unit test` is adding a schema in `/tests/interfaces`.

kaikai1024:
Would you please to check if @Kayryu is doing this? @leeyr338

Kayryu:
sorry! I didn't do it. If not urgent, I will add it

kaikai1024:
> sorry! I didn't do it. If not urgent, I will add it

You can try it if you would like to. I just need to know the progress and decide to do or not.

PR #379, opened 2019-03-29
Add test for getVersion

Kayryu:
Add test for getVersion in `/tests/interfaces.`

#### Applicable Issues
#359

### DIFF
--- a/tests/interfaces/rpc/tests/getVersion.json
+++ b/tests/interfaces/rpc/tests/getVersion.json
@@ -1,0 +1,30 @@
+{
+  "title": "getVersion",
+  "schema": {
+    "$ref": "../schemas/getVersion.json"
+  },
+  "tests": [
+    {
+      "title": "getVersion success if enabled",
+      "request": {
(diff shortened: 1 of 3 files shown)


tests/interfaces/rpc/tests/getVersion.json
```diff
@@ -0,0 +1,30 @@
...
+  "tests": [
+    {
+      "title": "getVersion success if enabled",
+      "request": {
+        "method": "getVersion",
+        "params": []
+      },
+      "expectedResponse": {
+        "result": {
+          "softwareVersion": "0.20"
```
kaikai1024: why here is `"0.20"`?

Kayryu: There is no specified version, it just a sample. If the software upgrade, the test needs to be changed.
Interface testing does not rely on specific versions.

kaikai1024: ok

Issue #385, opened 2019-04-01
Can not exit when run `bin/cita` after start the chain.


kaikai1024:
## Description

![Screenshot from 2019-04-01 15-31-36](https://user-images.githubusercontent.com/8768261/55310551-49682b00-5493-11e9-9f23-9b55b3877a04.png)

## Steps to Reproduce

1. Start the chain
2. Run `bin/cita`

**Expected behavior**: 

exit and back to the terminal.

**Actual behavior**: 

can not exit, even with `ctrl + c`

**Reproduce how often**: not sure(will try more times)

## Versions

develop

ouwenkg:
Well, I encounted the same situation with you.

kaikai1024:
because of the `bin/cita start test-chain/0` command.
`daemon`

u2:
> because of the `bin/cita start test-chain/0` command.
> `daemon`

Could you explain it more clearly? I did not get it.

kaikai1024:
> Could you explain it more clearly?

Still be finding the reason. daemon is just the situation.

`-d`  to use the docker then  `-i` can not exit.

yangby-cryptape:
`docker exec -it` will hang if there is any daemon process which hold the `stdout`.
(no reason for this, you can consider it is a bug or a feature)

We should redirect the `stdout` and `stderr` for `daemon processes`.

https://github.com/cryptape/cita/blob/5df437406fe4c885cbc3554004b254570b12b8c1/env.sh#L74-L80

PR #401, opened 2019-04-02
Redirect the stdout and stderr for daemon processes in docker.

yangby-cryptape:
### Summary

Fix: `docker exec -it` will hang if there is any daemon process which hold the `stdout` (#385).
(I find a issue about this: [moby/issue-33039](https://github.com/moby/moby/issues/33039#issuecomment-353250651))

### Further

This is a temporary solution, I think we should run a script in docker, instead of passing a shell command.

### Additional Knowledge: `"$@"` and `$*`

I think it's too difficult to explain this obscure concept.
So, I write a simple example to show it:

```bash
#!/usr/bin/env bash

function test_dollar () {
    echo "    Total: $# parameters(\$#)"
    echo "        parameter(\$0) = <$0>"
    local i=1
    while [ $# -ne 0 ]; do
        echo "        parameter(\$$i) = <$1>"
        shift 1
        i=$((i+1))
    done
}

echo "Print in function for \$@:"
test_dollar $@
echo "Print in function for \"\$@\":"
test_dollar "$@"
echo "Print in function for \"x y z \$@\":"
test_dollar "x y z $@"
echo "Print in function for \$*:"
test_dollar $*
echo "Print in function for \"\$*\":"
test_dollar "$*"
```

Run it and read the outputs. Have fun!

#### Explanation

Run `./dollar 1 "2 3" 4`:

- `$@` will split parameters by `$IFS` ([Internal Field Separator]), `"2 3"` will be consider as `2` and `3`, so output is:
  > Print in function for $@:
  >     Total: 4 parameters($#)
  >         parameter($0) = <./dollar>
  >         parameter($1) = <1>
  >         parameter($2) = <2>
  >         parameter($3) = <3>
  >         parameter($4) = <4>
- `"$@"` will split parameters as input, so output is:
  > Print in function for "$@":
  >     Total: 3 parameters($#)
  >         parameter($0) = <./dollar>
  >         parameter($1) = <1>
  >         parameter($2) = <2 3>
  >         parameter($3) = <4>
- `"x y z $@"` will do same things as `"$@"`, but `"x y z "` will be consider as prefix of the first parameter (in this example, `"1"` will be consider as `"x y z 1"`), so output is:
  > Print in function for "x y z $@":
  >     Total: 3 parameters($#)
  >         parameter($0) = <./dollar>
  >         parameter($1) = <x y z 1>
  >         parameter($2) = <2 3>
  >         parameter($3) = <4>
- `$*` is same as `$@`, so output is:
  > Print in function for $*:
  >     Total: 4 parameters($#)
  >         parameter($0) = <./dollar>
  >         parameter($1) = <1>
  >         parameter($2) = <2>
  >         parameter($3) = <3>
  >         parameter($4) = <4>
- `"$*"` will consider all inputs as a string, so output is: 
  > Print in function for "$*":
  >     Total: 1 parameters($#)
  >         parameter($0) = <./dollar>
  >         parameter($1) = <1 2 3 4>

So, in this PR, we have to use `$*` in `""`.

If we use `$@` as follow:

```bash
docker exec -d ${CONTAINER_NAME} /bin/bash -c "/usr/bin/gosu ${USER_NAME} $@ >/dev/null 2>&1"
```

When `$@` is `bin/cita start dev-chain/0`

The command will be parsed into 3 parts by shell: `"/usr/bin/gosu ${USER_NAME} bin/cita"`, `start` and `dev-chain/0`.
Actually, it should be 5 parts: `/usr/bin/gosu`, `${USER_NAME}`, `bin/cita`, `start` and `dev-chain/0`.

[Internal Field Separator]: https://en.wikipedia.org/wiki/Internal_field_separator

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -76,7 +76,7 @@ test -t 1 && USE_TTY="-t"
 # Condition `daemon` to run daemon.
 if [ "$3" == "--daemon"  ]; then
     set "${@:1:2}" "${@:4}"
-    docker exec -d ${CONTAINER_NAME} /usr/bin/gosu ${USER_NAME} "$@"
+    docker exec -d ${CONTAINER_NAME} /bin/bash -c "/usr/bin/gosu ${USER_NAME} $* >/dev/null 2>&1"
 elif [ $# -gt 0 ]; then
     docker exec -i ${USE_TTY} ${CONTAINER_NAME} /usr/bin/gosu ${USER_NAME} "$@"
 else


u2:
additional: https://stackoverflow.com/questions/2761723/what-is-the-difference-between-and-in-shell-scripts

u2:
Did not get the essence. :cry: 
```bash
➜  l2 bash test.sh 2 4
Print in function for $@:
    Total: 2 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2>
        parameter($2) = <4>
Print in function for "$@":
    Total: 2 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2>
        parameter($2) = <4>
Print in function for "x y z $@":
    Total: 2 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <x y z 2>
        parameter($2) = <4>
Print in function for $*:
    Total: 2 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2>
        parameter($2) = <4>
Print in function for "$*":
    Total: 1 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2 4>
➜  l2 bash test.sh '2 4'
Print in function for $@:
    Total: 2 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2>
        parameter($2) = <4>
Print in function for "$@":
    Total: 1 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2 4>
Print in function for "x y z $@":
    Total: 1 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <x y z 2 4>
Print in function for $*:
    Total: 2 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2>
        parameter($2) = <4>
Print in function for "$*":
    Total: 1 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <2 4>
➜  l2 bash test.sh 1 2 '3 4'
Print in function for $@:
    Total: 4 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <1>
        parameter($2) = <2>
        parameter($3) = <3>
        parameter($4) = <4>
Print in function for "$@":
    Total: 3 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <1>
        parameter($2) = <2>
        parameter($3) = <3 4>
Print in function for "x y z $@":
    Total: 3 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <x y z 1>
        parameter($2) = <2>
        parameter($3) = <3 4>
Print in function for $*:
    Total: 4 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <1>
        parameter($2) = <2>
        parameter($3) = <3>
        parameter($4) = <4>
Print in function for "$*":
    Total: 1 parameters($#)
        parameter($0) = <test.sh>
        parameter($1) = <1 2 3 4>
```

yangby-cryptape:
> Did not get the essence.

@u2 I updated the content of this PR.

kaikai1024:
> Actually, it should be 5 parts: /usr/bin/gosu, ${USER_NAME}, bin/cita, start and dev-chain/0

> "$*" will consider all inputs as a string, so output is:

"$*" is one parameter, should be five?

u2:
> > Actually, it should be 5 parts: /usr/bin/gosu, ${USER_NAME}, bin/cita, start and dev-chain/0
> 
> > "$*" will consider all inputs as a string, so output is:
> 
> "$*" is one parameter: `$1`, should be five?

The same question.

env.sh
```diff
@@ -76,7 +76,7 @@ test -t 1 && USE_TTY="-t"
 # Condition `daemon` to run daemon.
 if [ "$3" == "--daemon"  ]; then
     set "${@:1:2}" "${@:4}"
-    docker exec -d ${CONTAINER_NAME} /usr/bin/gosu ${USER_NAME} "$@"
+    docker exec -d ${CONTAINER_NAME} /bin/bash -c "/usr/bin/gosu ${USER_NAME} $* >/dev/null 2>&1"
```
leeyr338: User got the problem while they are *running CITA*, so `scripts/env.sh` is the exactly file needs to be changed?
And, we have a feature which need to print service's log to STDOUT, is ok for using `>/dev/null`?

Should we need to double check about this issue?

yangby-cryptape: As I said in the content of this PR:
> This is a temporary solution, I think we should run a script in docker, instead of passing a shell command.

And,
> `docker exec -it` will hang if there is any daemon process which hold the `stdout`.

So, we have no choice:
if you want to run a daemon process in docker and you want to have ability to use `docker exec -it` to attach into the docker environment, you **HAVE TO** redirect the `stdout`.

Yes, we can redirect `stdout` into a file, but, why not just keep it simple, do it just in the configuration of logs.

kaikai1024: > we have a feature which need to print service's log to STDOUT
Do not use `cita-env`, then the problem is solved

yangby-cryptape:
> 上面第一个引用部分说应该是五个部分，第二个引用部分说 "$*" 接受的输入都作为一个参数，是还会接着用空格进行分割还是我理解错了？

`$*` just outputs the raw parameters (separate by `$IFS`).
`"$*"` quote the all parameters, so it can be considered as only one parameter.

We pass `"/usr/bin/gosu ${USER_NAME} $* >/dev/null 2>&1"` to `/bin/bash -c`, `bash` will consider the whole parameter as a string, and `eval` it.

ouwenkg:
```
echo "Print in function for \"x y z \$*\":"
test_dollar "x y z $*"
```

```
Print in function for "x y z $*":
    Total: 1 parameters($#)
        parameter($0) = <./example.sh>
        parameter($1) = <x y z 1 2 3 4>
```

```
echo "Print in function for \"x y z \"\$*\"\":"
test_dollar "x y z "$*""
```

```
Print in function for "x y z "$*"":
    Total: 4 parameters($#)
        parameter($0) = <./example.sh>
        parameter($1) = <x y z 1>
        parameter($2) = <2>
        parameter($3) = <3>
        parameter($4) = <4>
```

Another two examples.

Issue #404, opened 2019-04-03
Add Editor Config 


kaikai1024:
## Description

whitespace or eof, etc.

## Additional Information

https://editorconfig.org/

PR #408, opened 2019-04-03
:scroll: Add editorconfig file. [skip ci]

kaikai1024:

A editorconfig file

#### Applicable Issues

#404

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,38 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+# Set default charset
+charset = utf-8
+indent_style = space
+indent_size = 4
+
+# Follow https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md
+[*.rs]
+indent_style = space
+indent_size = 4
+max_line_length = 100
+
+# Matches multiple files with brace expansion notation
+[*.{json,py,md,markdown}]
+indent_style = space
+indent_size = 4
+
+[Makefile]
+indent_style = tab
+indent_size = 4
+
+[*.yml]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,33 @@
...
+charset = utf-8
+indent_style = space
+indent_size = 4
+
+# Matches multiple files with brace expansion notation
+# Uncomment these when changing the default indent config
+# [*.{rs,json,py,md,markdown}]
+# indent_style = space
+# indent_size = 4
+
```
ouwenkg: why uncomment these lines?

ouwenkg: https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig

A little confused about your comments, explain more?

kaikai1024: Check this: https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties

Why i comment these lines?

check num 13-14:
```
indent_style = space
indent_size = 4
```

we already have the default indent config and it's same as the comment ones.
but if we change the indent config in the future(maybe indent_size = 2), while `{rs,json,py,md,markdown}` these should be stay the old config(indent_size =4).

If you config the indent on then `[]` part, use the config. if not use the default at the header of file.

ouwenkg: Ok，just a template confit usage.

kaikai1024: It's better to uncomment these in case of changing the default config. I will fix it later

.editorconfig
```diff
@@ -0,0 +1,38 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+# Set default charset
+charset = utf-8
+indent_style = space
+indent_size = 4
+
+# Follow https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md
```
yangby-cryptape: We already use CI (rust-fmt) to ensure this.

kaikai1024: yes.

It's better to have this or you need to run the rustFmt

PR #410, opened 2019-04-04
Use new bft

jerry-yu:
Fix not issue block, use new bft.


kaikai1024:
Better to fork the repo first.
Thx.

Issue #426, opened 2019-04-10
CITA p2p network High CPU usage


jiangxianliang007:
commit: 838bda448a6ac72ea78ab1257c27dae8711704f6

在32核心 64 G 机器上做性能测试， nentwork 打印大量的日志，且CPU 占比很高。最高的1200%，占了12核心。

![image](https://user-images.githubusercontent.com/24754263/55846243-7db8a700-5b77-11e9-9e49-858a770922f5.png)
![image](https://user-images.githubusercontent.com/24754263/55846399-2535d980-5b78-11e9-81cb-930374bbf433.png)

2019-04-10T10:01:06.330823947+08:00 - WARN - sub stream codec error: Custom { kind: InvalidData, error: FrameTooBig }
2019-04-10T10:01:06.330851496+08:00 - WARN - [P2pProtocol] Protocol Error, stream id: 3, protocol id: 1, error: IoError(Custom { kind: InvalidData, error: FrameTooBig })
2019-04-10T10:01:06.330871134+08:00 - WARN - [P2pProtocol] Protocol Error, stream id: 3, protocol id: 1, error: IoError(Custom { kind: InvalidData, error: FrameTooBig })
2019-04-10T10:01:06.330885619+08:00 - WARN - sub stream codec error: Custom { kind: InvalidData, error: FrameTooBig }
2019-04-10T10:01:06.330905249+08:00 - WARN - sub stream codec error: Custom { kind: InvalidData, error: FrameTooBig }
2019-04-10T10:01:06.330920854+08:00 - WARN - [P2pProtocol] Protocol Error, stream id: 3, protocol id: 1, error: IoError(Custom { kind: InvalidData, error: FrameTooBig })
2019-04-10T10:01:06.330933529+08:00 - WARN - [P2pProtocol] Protocol Error, stream id: 3, protocol id: 1, error: IoError(Custom { kind: InvalidData, error: FrameTooBig })
2019-04-10T10:01:06.330948510+08:00 - WARN - sub stream codec error: Custom { kind: InvalidData, error: FrameTooBig }
2019-04-10T10:01:06.330967508+08:00 - WARN - [P2pProtocol] Protocol Error, stream id: 3, protocol id: 1, error: IoError(Custom { kind: InvalidData, error: FrameTooBig })
2019-04-10T10:01:06.330983656+08:00 - WARN - sub stream codec error: Custom { kind: InvalidData, error: FrameTooBig }
[...]

PR #433, opened 2019-04-10
Bugfix: network high cpu usage

leeyr338:
Bugfix for network high cpu usage.
Issue: #426

### DIFF
--- a/cita-network/src/p2p_protocol/transfer.rs
+++ b/cita-network/src/p2p_protocol/transfer.rs
@@ -30,6 +30,11 @@ use tokio::codec::length_delimited::LengthDelimitedCodec;
 
 pub const TRANSFER_PROTOCOL_ID: ProtocolId = 1;
 
+// Quota (1 byte) = 200,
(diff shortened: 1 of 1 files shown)


cita-network/src/p2p_protocol/transfer.rs
```diff
@@ -29,6 +29,7 @@ use tentacle::{
 use tokio::codec::length_delimited::LengthDelimitedCodec;
 
 pub const TRANSFER_PROTOCOL_ID: ProtocolId = 1;
+pub const MAX_FRAME_LENGTH: usize = 64 * 1024 * 1204;
```
kaikai1024: why?: `64 * 1024 * 1204`

kaikai1024:
tips: You can use [skip travis]

cita-network/src/p2p_protocol/transfer.rs
leeyr338: Let MAX_FRAME_LENGTH = 512M for now. see #426

Issue #400, opened 2019-04-02
WARNING: Published ports are discarded when using host network mode


yangby-cryptape:
+ In linux, run `bin/cita start`:

  https://github.com/cryptape/cita/blob/5df437406fe4c885cbc3554004b254570b12b8c1/env.sh#L52-L62

  `--net=host` and `-p` are conflicted.

+ IMHO, use `host networking` as default present a potential security risk. 

Ref: [Docker Docs: Use host networking](https://docs.docker.com/network/host/)

kaikai1024:
> use host networking as default

That's a long history

kaikai1024:
> If you use the host network driver for a container, that container’s network stack is not isolated from the Docker host. 

> The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.


We should use defalue network mode: bridge

PR #443, opened 2019-04-11
Typo fixed in cita_config.sh.

clearloop:
## Typo
I found some typos in my previous pr, this pr contains three commits:

#### 1. [d4d1218]Fix config tools' typo.
This commit fixed the `config_tools.sh` —— an `fi` had been forgotten.

~~#### 2. [0fd5651]Adjusting format in env.sh.~~
This commit just optimize the format in `env.sh`, deletes some repeat lines.

~~#### 3. [8ca364b]Remove cargo files in docker container.~~
I'm not sure about this commit, docker volume invoked two wrong variables(This is my fault):
+ first time I try to replace them to the right variables. 
+ then I find, that If cargo files in docker can be removed directly?

### DIFF
--- a/scripts/cita_config.sh
+++ b/scripts/cita_config.sh
@@ -9,3 +9,4 @@ if [ -e $CITA_SCRIPTS/create_cita_config.py ]; then
     $CITA_SCRIPTS/create_cita_config.py $@
 else
     echo -e "\033[0;31mPlease run this command after build 🎨"
+fi


kaikai1024:
> If cargo files in docker can be removed directly?

More description about this? I'm a little confused.

env.sh
```diff
@@ -1,11 +1,17 @@
 #!/usr/bin/env bash
 
 if [ `uname` == 'Darwin' ]; then
+    SYSTEM_NET="bridge"
```
leeyr338: We don't need expose ports if we just use this script to build CITA.

env.sh
```diff
@@ -56,20 +45,16 @@ if [ $? -ne 0 ]; then
     docker run -d \
            --net=${SYSTEM_NET} \
            --volume ${SOURCE_DIR}:${WORKDIR} \
-           --volume ${DOCKER_CARGO}/git:${DOCKER_CARGO}/git \
```
leeyr338: This is a buffer for patch from crate.io. Do not delete this.

env.sh
kaikai1024: #400

env.sh
clearloop: Okay, I got it.

env.sh
```diff
@@ -28,14 +28,6 @@ USER_NAME="user"
 
 [[ "${USER_ID}" = "0" ]] && USER_NAME="root"
 
-# Init contanier's cargo, for logs.
-CARGO_HOME=/opt/.cargo
-DOCKER_CARGO=${HOME}/.docker_cargo
-mkdir -p ${DOCKER_CARGO}/git
-mkdir -p ${DOCKER_CARGO}/registry
-
-INIT_CMD="sleep infinity"
```
rainchen: should not delete this variable, it's very readable without a comment

env.sh
leeyr338: I mean we do not need to expose ports in `env.sh`, because this script just use to build `CITA`.
So we do not need this line, and 

```
 --net=${SYSTEM_NET} \

-p $EXPOSE ${DOCKER_IMAGE} \
```

clearloop: should I remove the `SYSTEM_NET` condition, and use `bridge` by default following #400 ?

kaikai1024: no.

we now have only one script.

https://github.com/cryptape/cita/blob/develop/scripts/release.sh#L58

kaikai1024: > Should I remove the `SYSTEM_NET` condition, and use `bridge` by default following #400 ?

Please do not do it in this pr.

leeyr338: Yeah, I have got this. Just forgot what i said.

kaikai1024:
tips: You can use `[skip travis]` if you don't change the system contracts.

Thx

Issue #366, opened 2019-03-26
 Integrate state test


ouwenkg:
## Description

Run [State test](https://github.com/ethereum/tests/blob/develop/GeneralStateTests/) in cita embedded state module.

## Record

* [x] ./tests/jsondata/GeneralStateTests/stRefundTest
* [x] ./tests/jsondata/GeneralStateTests/stCodeCopyTest
* [x] ./tests/jsondata/GeneralStateTests/stMemExpandingEIP150Calls
* [x] ./tests/jsondata/GeneralStateTests/stCallDelegateCodesCallCodeHomestead
* [x] ./tests/jsondata/GeneralStateTests/stRevertTest
* [x] ./tests/jsondata/GeneralStateTests/stArgsZeroOneBalance
* [x] ./tests/jsondata/GeneralStateTests/stMemoryStressTest
* [x] ./tests/jsondata/GeneralStateTests/stTransactionTest
* [x] ./tests/jsondata/GeneralStateTests/stReturnDataTest
* [x] ./tests/jsondata/GeneralStateTests/stPreCompiledContracts
* [x] ./tests/jsondata/GeneralStateTests/stCodeSizeLimit
* [x] ./tests/jsondata/GeneralStateTests/stHomesteadSpecific
* [x] ./tests/jsondata/GeneralStateTests/stEIP158Specific
* [x] ./tests/jsondata/GeneralStateTests/stZeroKnowledge2
* [x] ./tests/jsondata/GeneralStateTests/stCreateTest
* [x] ./tests/jsondata/GeneralStateTests/stSStoreTest
* [x] ./tests/jsondata/GeneralStateTests/stTransitionTest
* [x] ./tests/jsondata/GeneralStateTests/stZeroCallsTest
* [x] ./tests/jsondata/GeneralStateTests/stBadOpcode
* [x] ./tests/jsondata/GeneralStateTests/stLogTests
* [x] ./tests/jsondata/GeneralStateTests/stNonZeroCallsTest
* [x] ./tests/jsondata/GeneralStateTests/stCallDelegateCodesHomestead
* [x] ./tests/jsondata/GeneralStateTests/stBugs
* [x] ./tests/jsondata/GeneralStateTests/stShift
* [x] ./tests/jsondata/GeneralStateTests/stWalletTest
* [x] ./tests/jsondata/GeneralStateTests/stRandom2
* [x] ./tests/jsondata/GeneralStateTests/stEWASMTests
[...]

PR #457, opened 2019-04-12
Run state test

ouwenkg:
# Run state test

### Description of the Change

After fixing state tests in submodule, remove partial  skiped state tests.


kaikai1024:
nice.

Please update the issue #366

PR #469, opened 2019-04-16
ShellCheck env.sh

clearloop:
emmm, ShellCheck is a wonderful tool... 

refer to [shell-guide][1], modified:

+ `if [ true ]; then ...` -> `if [[ true ]]; then ...`
+  \`cmd\` -> `$(cmd)`
+ `$VAR` -> `"$VAR"`

[1]: https://github.com/cryptape/cita/blob/develop/style-guide/Shell.md

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -3,69 +3,65 @@
 # Environments' scripts.
 
 # OS
-if [[ $(uname) == 'Darwin' ]]; then
-    SYSTEM_NET="bridge"
-    SOURCE_DIR=$(dirname "$(realpath "$0")")
+if [[ "$(uname)" == 'Darwin' ]]; then
+    readonly SYSTEM_NET='bridge'
+    SOURCE_DIR="$(dirname "$(realpath "$0")")"
 else
-    SYSTEM_NET="host"
-    SOURCE_DIR=$(dirname "$(readlink -f "$0")")
+    readonly SYSTEM_NET="host"
+    SOURCE_DIR="$(dirname "$(readlink -f "$0")")"
 fi
 
 # Source Directory
-test -f "$SOURCE_DIR"/Cargo.toml
-ret="$?"
-if [[ "$ret" -eq 0 ]]; then
-    CONTAINER_NAME="cita_build_container"
-    DOCKER_IMAGE="cita/cita-build:ubuntu-18.04-20190304"
+if test -f "${SOURCE_DIR}/Cargo.toml"; then
+    readonly CONTAINER_NAME='cita_build_container'
+    readonly DOCKER_IMAGE='cita/cita-build:ubuntu-18.04-20190304'
 else
-    CONTAINER_NAME="cita_run_container"
-    DOCKER_IMAGE="cita/cita-run:ubuntu-18.04-20181009"
-    SOURCE_DIR=$(dirname "$SOURCE_DIR")
+    readonly CONTAINER_NAME='cita_run_container'
+    readonly DOCKER_IMAGE='cita/cita-run:ubuntu-18.04-20181009'
+    readonly SOURCE_DIR="$(dirname "$SOURCE_DIR")"
 fi
 
 # Patch from crates.io.
-CARGO_HOME=/opt/.cargo
-DOCKER_CARGO="$HOME"/.docker_cargo
-mkdir -p "$DOCKER_CARGO"/git
-mkdir -p "$DOCKER_CARGO"/registry
+readonly CARGO_HOME='/opt/.cargo'
+readonly DOCKER_CARGO="${HOME}/.docker_cargo"
+mkdir -p "${DOCKER_CARGO}/git"
+mkdir -p "${DOCKER_CARGO}/registry"
 
 # Docker Port
-EXPOSE="1337:1337"
+EXPOSE='1337:1337'
 if [[ "$3" == "port" ]]; then
-    EXPOSE="${*:4}"
-    [[ "$EXPOSE" = "" ]] && EXPOSE="1337:1337"
-    docker container stop $CONTAINER_NAME > /dev/null 2>&1
-    echo -e "\033[0;32mExpose ports: ${*}\033[0m"
+    EXPOSE=( "${@:4}" )
+    [[ "${EXPOSE[*]}" == "" ]] && EXPOSE=('1337:1337')
+    docker container stop "${CONTAINER_NAME}" > /dev/null 2>&1
+    echo -e "\033[0;32mExpose ports: ${EXPOSE[*]} \033[0m"
 fi
 
 # Docker Arguments
-WORKDIR=/opt/cita
 USER_ID="$(id -u "$USER")"
-USER_NAME="user"
+readonly WORKDIR='/opt/cita'
+readonly USER_NAME='user'
 
-cp /etc/localtime "$SOURCE_DIR"/localtime
-LOCALTIME_PATH="$SOURCE_DIR"/localtime
-[[ ${USER_ID} = "0" ]] && USER_NAME="root"
+cp '/etc/localtime' "${SOURCE_DIR}/localtime"
+readonly LOCALTIME_PATH="${SOURCE_DIR}/localtime"
+[[ "${USER_ID}" = '0' ]] && USER_NAME='root'
 
-INIT_CMD="sleep infinity"
+readonly INIT_CMD='sleep infinity'
 
 # Run Docker
-docker ps | grep ${CONTAINER_NAME} > /dev/null 2>&1
-ret="$?"
-if [[ "$ret" -ne 0 ]]; then
-    echo "Start docker container $CONTAINER_NAME ..."
-    docker rm ${CONTAINER_NAME} > /dev/null 2>&1
+if ! docker ps | grep "${CONTAINER_NAME}" > '/dev/null' 2>&1; then
+    echo "Start docker container ${CONTAINER_NAME} ..."
+    docker rm "${CONTAINER_NAME}" > '/dev/null' 2>&1
     docker run -d \
-           --net=${SYSTEM_NET} \
-           --volume "$SOURCE_DIR":"$WORKDIR" \
-           --volume "$DOCKER_CARGO"/git:"$CARGO_HOME"/git \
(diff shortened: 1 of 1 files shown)


env.sh
```diff
@@ -1,87 +1,85 @@
...
+    SOURCE_DIR="$(dirname "$SOURCE_DIR")"
 fi
 
-if [ `uname` == 'Darwin' ]; then
-    cp /etc/localtime ${SOURCE_DIR}/localtime
-    SYSTEM_NET="bridge"
-    LOCALTIME_PATH="${SOURCE_DIR}/localtime"
-else
-    SYSTEM_NET="host"
-    LOCALTIME_PATH="/etc/localtime"
```
clearloop: + Merged this condition to `line-4`
+ LOCALTIME_PATH move to `line-46`

env.sh
```diff
@@ -1,87 +1,85 @@
...
 [[ "$3" == "start" ]] || [[ "$3" == "restart" ]] && set "${@:1:2}" "--daemon" "${@:3}"
 
-# Condition `daemon` to run daemon.
-if [ "$3" == "--daemon"  ]; then
+if [[ "$3" == "--daemon" ]]; then
     set "${@:1:2}" "${@:4}"
-    docker exec -d ${CONTAINER_NAME} /bin/bash -c "/usr/bin/gosu ${USER_NAME} $* >/dev/null 2>&1"
-elif [ $# -gt 0 ]; then
+    docker exec -d ${CONTAINER_NAME} /bin/bash -c /usr/bin/gosu ${USER_NAME} "$@" >/dev/null 2>&1
+elif [[ $# -gt 0 ]]; then
```
ouwenkg: I am not sure about the correct usage `$@` or `$*`. But，we had changed `$@` to `$*` in #401 on purpose.

@yangby-cryptape  Sorry to take you some times to review this command again.

kaikai1024: I can't start the chain...

clearloop: `ShellCheck` recommends to use `$*` too, as it [says][1]: 

> Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

+ `$*` for `string`
+ `$@` for separated arguments

I guess.

I used to use `$@` and never using `$*` so I didn't change it in case...

This modification was written last week and I forget to pr XD, so I didn't notice the new change.

The explanation in #401 does make sense, I'll change `$@` to `$*` in docker commands and do more test about it.

[1]: https://github.com/koalaman/shellcheck/wiki/SC2124

clearloop: > Maybe it's better to run the travis that uses `env.sh`.
> And the circleCI uses `bebop`

I like modifying shell commands, and I have to try to test them comprehensively(both installing and running), because the shell scripts are the entrance to cita users, every little change might create big problems to users using cita.

I find I depended on travis overly after I noticed them in my first pr, even though I don't know exactly if travis' scripts covered every tests about my changing.

I always modifying too many lines in shell scripts though I feel stressed with every pr I used to pr...but anyway, I should make `env.sh` and `scripts/cita` performing best in where I had modified.

kaikai1024: > every little change might create big problems to users using cita.

That's our fault. we should add test of cita usage into CI

> I always modifying too many lines in shell scripts though I feel stressed with every pr I made

Small pr, happy life :sunglasses:

clearloop: Just ignore my ranting 🤦🏻‍♂️

> Small pr, happy life 😎

Might be the truth...

> I can't start the chain...

Just catch the bug,  as #401's patch: 

```shell
# ATTENTION: `-d` required passing full string.                                                                                                                                                           
docker exec -d ${CONTAINER_NAME} /bin/bash -c "/usr/bin/gosu ${USER_NAME} $* >/dev/null 2>&1"
```

+ `docker exec -d` should passing full string: `$*`, because it crosses two tty? and docker does special processes to the `string`? 
+ but `docker exec -i` should use parameters: `$@`, because of the normal passing?

Really a big discovery for me, I used to test `docker exec -d` over and over again in my early pr, and never noticed this question.

env.sh
```diff
@@ -3,84 +3,83 @@
 # Environments' scripts.
 
 # OS
-if [[ $(uname) == 'Darwin' ]]; then
-    SYSTEM_NET="bridge"
-    SOURCE_DIR=$(dirname "$(realpath "$0")")
+if [[ "$(uname)" == 'Darwin' ]]; then
+    readonly SYSTEM_NET='bridge'
```
clearloop: [Readonly variables][1] in google shell-style guide.

[1]: https://google.github.io/styleguide/shell.xml?showone=Read-only_Variables#Read-only_Variables

env.sh
```diff
@@ -3,84 +3,83 @@
 # Environments' scripts.
 
 # OS
-if [[ $(uname) == 'Darwin' ]]; then
-    SYSTEM_NET="bridge"
-    SOURCE_DIR=$(dirname "$(realpath "$0")")
+if [[ "$(uname)" == 'Darwin' ]]; then
```
clearloop: 1. [Test, \[ and \[\[][1]: `[[ ... ]]`
2. [Command Substitution][2]: `"$(command)"`
3. [Single Quoting][3]: `'Darwin'`

[1]: https://google.github.io/styleguide/shell.xml?showone=Quoting#Test,_%5B_and_%5B%5B
[2]: https://google.github.io/styleguide/shell.xml?showone=Command_Substitution#Command_Substitution
[3]: https://google.github.io/styleguide/shell.xml?showone=Quoting#Quoting

env.sh
```diff
@@ -3,84 +3,83 @@
...
 
 # Docker Port
-EXPOSE="1337:1337"
+EXPOSE='1337:1337'
 if [[ "$3" == "port" ]]; then
-    EXPOSE="${*:4}"
-    [[ "$EXPOSE" = "" ]] && EXPOSE="1337:1337"
-    docker container stop $CONTAINER_NAME > /dev/null 2>&1
-    echo -e "\033[0;32mExpose ports: ${*}\033[0m"
+    EXPOSE=( "${@:4}" )
```
clearloop: 5. [SC2128][5]: `myarray=(foo bar)`

[5]: https://github.com/koalaman/shellcheck/wiki/SC2128

env.sh
```diff
@@ -3,84 +3,83 @@
...
-INIT_CMD="sleep infinity"
+readonly INIT_CMD='sleep infinity'
 
 # Run Docker
-docker ps | grep ${CONTAINER_NAME} > /dev/null 2>&1
-ret="$?"
-if [[ "$ret" -ne 0 ]]; then
-    echo "Start docker container $CONTAINER_NAME ..."
-    docker rm ${CONTAINER_NAME} > /dev/null 2>&1
+if ! docker ps | grep "${CONTAINER_NAME}" > '/dev/null' 2>&1; then
```
clearloop: 6. [SC2182][6]: `if ! make mytarget`

[6]: https://github.com/koalaman/shellcheck/wiki/SC2128

env.sh
```diff
@@ -3,84 +3,83 @@
...
+[[ "$3" == 'start' ]] || [[ "$3" == 'restart' ]] && set "${@:1:2}" '--daemon' "${@:3}"
 
-if [[ "$3" == "--daemon" ]]; then
+if [[ "$3" == '--daemon' ]]; then
     set "${@:1:2}" "${@:4}"
-    docker exec -d ${CONTAINER_NAME} /bin/bash -c /usr/bin/gosu ${USER_NAME} "$@" >/dev/null 2>&1
+    # ATTENTION:
+    # (docker commands) (parameters) /bin/bash -c (full-string docker arguments || sepreate them on by one)
+    # `bin/bash -c` is local commands.
+    docker exec -d "${CONTAINER_NAME}" /bin/bash -c "/usr/bin/gosu ${USER_NAME} ${*} >/dev/null 2>&1"
```
clearloop: 7. [SC2145][7]: `$*`

[7]: https://github.com/koalaman/shellcheck/wiki/SC2145

yangby-cryptape:
I know, the old code is a piece of :shit:.

But, do not use so many `readonly` `global` variables.

Define functions and pass `local` variables between functions will make it better and more flexible.

env.sh
```diff
@@ -1,88 +1,85 @@
...
-if [ `uname` == 'Darwin' ]; then
-    SOURCE_DIR="$(dirname $(realpath $0))"
+# OS
+if [[ "$(uname)" == 'Darwin' ]]; then
+    readonly SYSTEM_NET='bridge'
+    SOURCE_DIR="$(dirname "$(realpath "$0")")"
 else
-    SOURCE_DIR="$(dirname $(readlink -f $0))"
+    readonly SYSTEM_NET="host"
+    SOURCE_DIR="$(dirname "$(readlink -f "$0")")"
```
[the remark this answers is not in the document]
clearloop: If we symlink `cita` in a global path, just like:

```
/usr/local/bin/cita
```

using `$0`, the `SOURCE_DIR` will point to `/usr/local/bin`

clearloop:
> Define functions and pass local variables between functions will make it better and more flexible.

Yep,  wrapping scripts into functions is a good idea, but I can't change more lines that are not about formats in the pr, it will make the pr more complex XD

And, based on the `$0` question, I just want to optimize the two [conditions][1] about `Darwin` and `Linux` into one actually...while I saw [cita-shell-style-guide][1] in my last pr,  trying to apply the formats in this file.

https://github.com/cryptape/cita/blob/7049a5ac115a5446c98ea27f5ee9faab5e4acfc3/env.sh#L3-L7

https://github.com/cryptape/cita/blob/7049a5ac115a5446c98ea27f5ee9faab5e4acfc3/env.sh#L19-L26

[1]: https://github.com/cryptape/cita/blob/develop/style-guide/Shell.md

env.sh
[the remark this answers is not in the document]
clearloop:
```
# Mac Only? In my Linux, it works.
SOURCE_DIR="$(dirname "$(realpath "$0")")"
```
I stuck this on my first pr —— `travis` might not like the `realpath` function.

kaikai1024:
>  it will make the pr more complex

yeah, @yangby-cryptape 's advice should be another pr.

env.sh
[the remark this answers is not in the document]
kaikai1024: What do you think of the reply? @yangby-cryptape

PR #484, opened 2019-04-18
Update Rust toolchain to `1.34.0`.

kaikai1024:
### Description of the Change

Update Rust toolchain to `1.34.0`.

* [ ] TryInto & TryFrom *Maybe nex pr*

### Applicable Issues

#458

Issue #458, opened 2019-04-13
New image


kaikai1024:
## Description

* [x] #420 
* [ ] #444
* [x] #455 
* [ ] ~~#456~~
* [ ] ~~Add toolchain file to keep the rust version be the same with image. TBD~~

PR #489, opened 2019-04-19
Update Rust toolchain to `1.34.0`.

kaikai1024:
### Description of the Change

Update Rust toolchain to 1.34.0.

TODO:  Replace TryInto & TryFrom in nex pr

Closed pr: #484

### Applicable Issues

#458

### DIFF
--- a/cita-executor/core/src/state/account.rs
+++ b/cita-executor/core/src/state/account.rs
@@ -28,6 +28,7 @@ use pod_account::*;
 use rlp::*;
 use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
+use std::convert::Into;
 use std::fmt;
 use std::sync::Arc;
 use types::basic_account::BasicAccount;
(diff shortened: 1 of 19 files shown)


cita-executor/core/src/state/account.rs
```diff
@@ -271,7 +272,7 @@ impl Account {
     /// Verify value proof of the trie's storage at `key`.
     pub fn verify_value_proof(&self, key: &H256, proof: &[Bytes]) -> Option<H256> {
         trie::triedb::verify_value_proof(key, self.storage_root, proof, ::rlp::decode)
-            .map(|v: U256| v.into())
+            .map(&Into::into as &Fn(U256) -> H256)
```
yangby-cryptape: I guess `&Into::into as &Fn(_) -> H256` is ok, too.

kaikai1024: I can't try now. will check it later.
But why not `&Into::into as &Fn(U256) -> _`?

yangby-cryptape: The input of `Into::into` is determined by `verify_value_proof(..)`.
So rust compiler can know this **implicitly**.

But rust compiler can't know the type of output.
So you have to tell rust compiler **explicitly**.

In fact, if the rust compiler check the function signature, I think it can know the type of output.
Maybe the output `Option<H256>` is **out of the scope** which it uses to determine the type.

Example:
```rust
fn f<T>(x: Vec<T>) -> Vec<T> { x }
fn g<T>(x: T) -> T { x }
fn main() {
    // failed
    let _ = (1..100).collect();
    // passed
    let _ = (1..100).collect::<Vec<_>>();
    // failed
    let t = (1..100).collect();
    let _ = g(t);
    // passed
    let t = (1..100).collect();
    let t = g(t);
    let _ = f(t);

}
```

PR #495, opened 2019-04-22
Modify script for new network config. [skip travis]

leeyr338:
Modify script for new network config:
- Do not need `ca` files;
- Do not need `id_card`;

### DIFF
--- a/scripts/create_cita_config.py
+++ b/scripts/create_cita_config.py
@@ -162,11 +162,7 @@ class ChainInfo():
                                              'authorities.list')
         self.nodes = NetworkAddressList()
         self.enable_tls = False
-        self.root_ca_name = 'rootCA'
-        self.node_ca_name = chain_name
-        self.server_ca_name = 'server'
         self.prefix_subj = '/C=CN/ST=ZJ/O=Cryptape, Inc./CN='
-        self.ca_days = 36525
 
     def template_create_from_arguments(self, args, contracts_dir_src,
                                        configs_dir_src):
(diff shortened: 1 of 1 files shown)


scripts/create_cita_config.py
```diff
@@ -400,39 +340,40 @@ def run_subcmd_create(args, work_dir):
...
+                                       title='subcommands',
+                                       help='additional help')
 
     #
     # Subcommand: create
     #
 
-    pcreate = subparsers.add_parser(
-        SUBCMD_CREATE, help='create a chain with nodes')
+    pcreate = subparsers.add_parser(SUBCMD_CREATE,
```
kaikai1024: auto-fmt?pylint?

leeyr338: I use command:
```shell
yapf -i create_cita_config.py
```
and yapf version is `0.27.0`.

Am I make something wrong?

kaikai1024: I see.
Why use `yapf`? and it should be another commit if you format the code.

scripts/create_cita_config.py
```diff
@@ -479,33 +421,31 @@ def parse_arguments():
...
-        type=NetworkAddressList.from_str_get_one,
-        metavar='IP:PORT',
-        help='Node network addresses for new nodes.')
+    pappend.add_argument('--node',
+                         required=True,
+                         type=NetworkAddressList.from_str_get_one,
+                         metavar='IP:PORT',
+                         help='Node network addresses for new nodes.')
 
     pappend.add_argument(
```
kaikai1024: Why is this line not formatted?

scripts/create_cita_config.py
leeyr338: Oh, `pylint` is okay!

scripts/create_cita_config.py
leeyr338: I don't know why, maybe `yapf` makes something wrong. I will use `pylint`.

scripts/create_cita_config.py
kaikai1024: > For instance, code that conforms to the PEP 8 guidelines may not be reformatted. But it doesn't mean that the code looks good.

> It's based off of 'clang-format'

scripts/create_cita_config.py
kaikai1024: You can check it on next pr. use yapf or not

scripts/create_cita_config.py
```diff
@@ -451,7 +390,7 @@ def parse_arguments():
         default=KeyKeyValueDict(),
         metavar='Contract.Argument=Value',
         help='Update constructor arguments for system contract.'
-        ' Can be specify more than once.')
+             ' Can be specify more than once.')
```
kaikai1024: this

scripts/create_cita_config.py
```diff
@@ -23,8 +23,8 @@ def update_search_paths(work_dir):
 
 def generate_prevhash(resource_dir):
     if resource_dir is not None \
-            and os.path.exists(resource_dir) \
-            and os.path.isdir(resource_dir):
+        and os.path.exists(resource_dir) \
```
kaikai1024: no need to change

Issue #494, opened 2019-04-22
Node doesn't work well after get rpc request getLogs with a big param of toBlock.


luqz:
## Description

Node doesn't work well after get rpc request getLogs with a big param of to.

## Steps to Reproduce

1. create and start a test-chain;
2. use rpc method getLogs, and param `toBlock ` is very big, like 0xffffffffffffff;

**Expected behavior**: 
1.  Return array or error message, 
2. The node works well.

**Actual behavior**: 
1.  Return error message: system timeout, 
2. Module cita-chain keep use more than 90% of CPU,
3. The node almost doesn't work.

**Reproduce how often**: 
every time.

## Versions

v0.22.0

leeyr338:
>> use rpc method getLogs, and param to is very big, like 0xffffffffffffff;

This is no `to` param for `getLogs`, see
https://docs.citahub.com/zh-CN/cita/rpc-guide/rpc#getlogs

Would you please show me the exactly `curl` command for your JSON-RPC call.

leeyr338:
From @jerry-yu 
jsonrpc的getlogs接口，可能获取的logs可能会超级大。
logs编码设置进入response，占用chain很多内存，
可能会导致内存不足.

另外还有一个思考就是：业务发送new filterlog，然后chain重启了，业务层感知不到，没有继续发送new filterlog，可能就没法准确的获得 change logs了

kaikai1024:
> This is no to param for getLogs,

it does have.

This is no to param for getLogs,

kaikai1024:
Does it hapen if `to` is `latest`?

leeyr338:
> Does it hapen if to is latest?

It works for latest.

mingxiaowu:
v0.23 验证通过

PR #499, opened 2019-04-23
Fix: getLogs break down the chain when toBlock very large.

leeyr338:
Fix: getLogs break down the chain when toBlock very large.

```rust
let end_block = if to_block > pending_block { pending_block } else { to_block };
```

Issue:
#494


leeyr338:
> Please add the description about this into the citahub-docs later.

OK.

kaikai1024:
Please cherry-pick the commit into the `release-v0.23.0` @KaoImin

Issue #496, opened 2019-04-23
snapshot don't take effect


QingYanL:
## Description

[Description of the issue]
snapshot don't take effect

## Steps to Reproduce

1.  **exec in mac**   
`cd /test-chain/0`
`../../bin/snapshot_tool -m snapshot -e 100`
output: **-bash: ../../bin/snapshot_tool: cannot execute binary file**
2. `ls -l snapshot_tool`
output: **-rwxr-xr-x@ 1 yanli  staff  9901816  4 19 01:36 snapshot_tool**

![image](https://user-images.githubusercontent.com/48231505/56551965-540e6f80-65bc-11e9-8c22-21a91fc1a9dd.png)

3. **exec in docker**   
`cd /test-chain/0`
`../../bin/snapshot_tool  -m snapshot -e 100`
`ls snap*`
output:  **ls: cannot access 'snap': No such file or directory**

![image](https://user-images.githubusercontent.com/48231505/56552013-77d1b580-65bc-11e9-88cf-be1880f89cd5.png)

**Expected behavior**: [snapshot take effect]

**Actual behavior**: [snapshot don't take effect]

**Reproduce how often**: [every time]

## Versions

[develop]

ouwenkg:
Two problems:
1. `-bash: ../../bin/snapshot_tool: cannot execute binary file`, it only occured in mac os. Platform difference causes this error. It's recommend to run such command in docker.
2. `ls snap*` and return no result. That a bug in snapshot module, the process will panic if we use default file path. #500  will fix it.

@QingYanL Check it.

PR #500, opened 2019-04-23
fix panic in snapshot path

ouwenkg:
# Fix panic in snapshot path

选区_037

### Applicable Issues
#496

### DIFF
--- a/tools/snapshot_tool/src/main.rs
+++ b/tools/snapshot_tool/src/main.rs
@@ -58,9 +58,7 @@ fn main() {
         .get_matches();
     let command = matches.value_of("cmd").expect("provide specific command");
     let command = cast_command(command);
-    let file = matches
-        .value_of("file")
-        .expect("provice specific snapshot-file path");
+    let file = matches.value_of("file").unwrap_or("snapshot");
     let file = file.to_owned();
     let start_height = matches.value_of("start_height").unwrap_or("0");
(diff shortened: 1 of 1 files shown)


tools/snapshot_tool/src/main.rs
```diff
@@ -58,9 +58,7 @@ fn main() {
         .get_matches();
     let command = matches.value_of("cmd").expect("provide specific command");
     let command = cast_command(command);
-    let file = matches
-        .value_of("file")
-        .expect("provice specific snapshot-file path");
+    let file = matches.value_of("file").unwrap_or("snapshot");
```
leeyr338: User get error message:
```
output: -bash: ../../bin/snapshot_tool: cannot execute binary file
```
Why this change can fix the bug? Has user run `panic` ?

ouwenkg: Noop! 
This change solves `ls snap*` no result problem. Using default file path config, snapshot tool will panic.

`cannot execute binary file` it occured only in mac，caused by different platforms.

kaikai1024: So it's nothing about the #496

> cannot execute binary file it occured only in mac，caused by different platforms.

Update the conclusion into the #496.

kaikai1024:
@KaoImin And this commit.

Issue #493, opened 2019-04-22
block.coinbase can't be used in autoExec()


luqz:
## Description

block.coinbase can't be used in autoExec()

## Steps to Reproduce

1. Add `SysConfig.autoExec=true` when create chain;
2. use `address(block.coinbase).transfer(1)` in function `autoExec()`;
3. create contract and registe it to AutoExec contract;
4. Transfer token to contract.

**Expected behavior**: contract balance sub 1  per block

**Actual behavior**: contract balance remain unchanged

**Reproduce how often**: every time.

## Versions

v0.22.0

## Additional Information

`address(0x5444B13eFe73A7de4d56f2c043424115C8e26eF1).transfer(1)` works but `address(block.coinbase).transfer(1)` not.

kaikai1024:
What's the `Solidity` version?

luqz:
> What's the `Solidity` version?

^0.4.24

kaikai1024:
I've tried the `block.coinbase` on a normal conract and it works.
Maybe something wrong with the auto-exec. 

#483

kaikai1024:
Would you give some other info? @luqz  I can not reproduce.

the contract:

```
pragma solidity ^0.4.24;

contract X {
    address public addr;
    uint public x;

    constructor() payable public {
        addr = block.coinbase;
    }

    function autoExec() public {
        address(block.coinbase).transfer(1);
        x++;
    }
}
```

https://gist.github.com/kaikai1024/74944724904174cd59e4116dd6456b6e

luqz:
Sorry, I try it again.  Your code works well.
I found that maybe the fact is block.coinbase==0x0.
If I remove `require(to != address(0), "Invalid Address")`, It works, but not Expected.
My contract:
`
pragma solidity ^0.4.24;

import './IAutoExec.sol';

contract TestAutoExec is IAutoExec {
    uint public count;

    event Transfer(address indexed from, address indexed to, uint256 indexed value);

    modifier onlyAutoExec {
        if (msg.sender == address(0xFFFfFfFffffffFfFFFFffffFfFFFffFFfF020013))
            _;
        else return;
    }

    function autoExec() external onlyAutoExec {
        incentive();
    }


    function incentive() public {
        _transfer(address(block.coinbase), 1);
        count ++;
    }


    function _transfer(address to, uint value) private returns (bool) {
        require(address(this).balance >= value, "Balance not enough");
        require(to != address(0), "Invalid Address");
        to.transfer(value);
[...]

PR #506, opened 2019-04-25
fix coinbase in auto exec

ouwenkg:
# fix coinbase in auto exec

### Description of the Change

In the case of auto exec situation, coinbase should be proposer.

### Applicable Issues
#493

### DIFF
--- a/cita-executor/core/src/libexecutor/auto_exec.rs
+++ b/cita-executor/core/src/libexecutor/auto_exec.rs
@@ -41,11 +41,12 @@ lazy_static! {
     static ref AUTO_EXEC_HASH: Vec<u8> = method_tools::encode_to_vec(AUTO_EXEC);
 }
 
-// pub fn auto_exec(state: &mut State<StateDB>) -> evm::Result<FinalizationResult> {
 pub fn auto_exec(
     state: &mut State<StateDB>,
     auto_exec_quota_limit: u64,
     economical_model: EconomicalModel,
+    // TODO: UPGRADE PROTOC0L VERSION
+    env_info: EnvInfo,
 ) {
     let hash = &*AUTO_EXEC_HASH;
     let params = ActionParams {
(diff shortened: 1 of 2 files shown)


cita-executor/core/src/libexecutor/auto_exec.rs
```diff
@@ -68,7 +69,7 @@ pub fn auto_exec(
     let mut trace_output = tracer.prepare_trace_output();
     let output = OutputPolicy::Return(BytesRef::Flexible(&mut out), trace_output.as_mut());
     let factory = Factory::new(VMType::Interpreter, 1024 * 32);
-    let env_info = EnvInfo::default();
```
kaikai1024: Just delete it.

cita-executor/core/src/libexecutor/block.rs
```diff
@@ -208,6 +206,9 @@ impl ExecutedBlock {
                 b
             });
 
+        block.set_state_root(state_root);
```
kaikai1024: It's nothing about the commit.
A new commit is better.

Actually it's not a good habit to do things beyond the pr's purpose.
I've done this before, too. Maybe we all should follow it and review this each other.

kaikai1024: BTW, I'll write the doc about the rule: PR guide, review guide, and something like this.

cita-executor/core/src/libexecutor/auto_exec.rs
```diff
@@ -41,11 +41,12 @@ lazy_static! {
     static ref AUTO_EXEC_HASH: Vec<u8> = method_tools::encode_to_vec(AUTO_EXEC);
 }
 
-// pub fn auto_exec(state: &mut State<StateDB>) -> evm::Result<FinalizationResult> {
 pub fn auto_exec(
     state: &mut State<StateDB>,
     auto_exec_quota_limit: u64,
     economical_model: EconomicalModel,
+    // TODO: UPGRADE PROTOCAL VERSION
```
kaikai1024: `PROTOCAL` -> `PROTOCOL`

cita-executor/core/src/libexecutor/auto_exec.rs
ouwenkg: Done!

cita-executor/core/src/libexecutor/auto_exec.rs
ouwenkg: Deal!

cita-executor/core/src/libexecutor/auto_exec.rs
kaikai1024: I mean the blank line.
That's ok.

PR #510, opened 2019-04-25
update log in forward

ouwenkg:
### Description of the Change

Update partial logs print format


ouwenkg:
> cool.
> 
> I remember we have the `[]` and `:` format.
> We need a unified log rule.

Indeed !!!

PR #533, opened 2019-04-30
Bump rust toolchain to `1.34.1`.

kaikai1024:
### Description of the Change

Rust version [`1.34.1`](https://blog.rust-lang.org/2019/04/25/Rust-1.34.1.html)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ alias:
 
   - &job-default
     docker:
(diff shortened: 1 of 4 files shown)


leeyr338:
Why  we need to do this?

kaikai1024:
> Why we need to do this?

We will always follow the latest release version of Rust *unless there are special reasons*.

PR #537, opened 2019-05-04
Upgrade the packages to avoid the security alert.

kaikai1024:
### Description of the Change

* Upgrde all the packages

Actually the alert is because of the [cita-skd-js](https://github.com/cryptape/cita-sdk-js/network/alerts):

* `0.21.1` of `cryptape/cita-sdk`
* `2.2.2` of `cryptape/cita-signer`


kaikai1024:
`skip circle` does not work. check [this](https://discuss.circleci.com/t/adding-circle-skip-to-commit-message-to-skip/2778/19)

PR #574, opened 2019-05-15
add discovery flag to network.toml

jerry-yu:
add discovery flag to network.toml

### DIFF
--- a/scripts/create_cita_config.py
+++ b/scripts/create_cita_config.py
@@ -162,6 +162,7 @@ class ChainInfo():
                                              'authorities.list')
         self.nodes = NetworkAddressList()
         self.enable_tls = False
+        self.enable_discovery = False
         self.prefix_subj = '/C=CN/ST=ZJ/O=Cryptape, Inc./CN='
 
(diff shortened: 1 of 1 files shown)


scripts/create_cita_config.py
```diff
@@ -417,6 +419,12 @@ def parse_arguments():
         action='store_true',
         help='The data is encrypted and transmitted on the network')
 
+    # enable encrypted
```
kaikai1024: enable discovery

scripts/create_cita_config.py
```diff
@@ -162,6 +162,7 @@ def __init__(self, chain_name, output_dir):
                                              'authorities.list')
         self.nodes = NetworkAddressList()
         self.enable_tls = False
+        self.enable_discovery = False
```
leeyr338: I think it better set `enable discovery` as default.

leeyr338: And this will cause `discovery test` failed.

jerry-yu:
there maybe some proplem when enable disvovery

Issue #477, opened 2019-04-18
suicide refund problem


ouwenkg:
## suicide refund problem

[Description of the issue]

In CITA:
```
 fn suicide(&mut self, refund_address: &Address) -> evm::Result<()> {
        if self.static_flag {
            return Err(evm::Error::MutableCallInStaticContext);
        }

        let address = self.origin_info.address;
        let balance = self.balance(&address)?;
        self.tracer.trace_suicide(address, balance, *refund_address);
        self.substate.suicides.insert(address);
        Ok(())
    }
```
In Ethereum:
```
 fn suicide(&mut self, refund_address: &Address) -> evm::Result<()> {
        if self.static_flag {
            return Err(evm::Error::MutableCallInStaticContext);
        }

        let address = self.origin_info.address;
        let balance = self.balance(&address)?;
       
        if &address == refund_address {
            self.state.sub_balance(&address, &balance)?;
        } else {
            self.state
                .transfer_balance(&address, refund_address, &balance)?;
        }

        self.tracer.trace_suicide(address, balance, *refund_address);
        self.substate.suicides.insert(address);
        Ok(())
    }
```

CITA did nothing about self-destruct refund process.

## Additional Information
#366 #418

PR #576, opened 2019-05-15
Add coinbase and suicide

kaikai1024:
### Description of the Change

Add test case:

* Get coinbase when use auto-exec
* Refund token when suicide

### Applicable Issues

#477 #493


kaikai1024:
Can not trigger travisCI....

PR #585, opened 2019-05-16
fix bug for save wrong session id

jerry-yu:
fix bug for save wrong session id


kaikai1024:
Please add some description in the commit, like:

> Fix bug for saving wrong session id

kaikai1024:
@leeyr338 You are in charge of this.

PR #608, opened 2019-05-27
Update cita-bft submodule

rev-chaos:
### Description of the Change
Refractor cita-bft project, dependent on consensus libary - bft-rs.


kaikai1024:
Should be PR of cita-bft first.

Issue #604, opened 2019-05-24
Fix the Circle CI


kaikai1024:
## Description

* [x] `cita_snapshot_test`
* [x] `cita_tls_basic.sh`

https://circleci.com/gh/cryptape/cita/7413

## Additional Information

#603

kaikai1024:
Occasionally...
Maybe the timeout.

Reopen it when it happened again.

kaikai1024:
Rerun from the failed is ok.

kaikai1024:
https://circleci.com/workflow-run/76a32d1f-764b-4cc1-a790-536e345ae142

kaikai1024:
https://circleci.com/gh/cryptape/cita/7714

PR #610, opened 2019-05-28
Use cita-logger crate. [skip travis]

kaikai1024:
### Description of the Change

* update cita-bft
* update cita-forever
* update cita-common
* use cita-logger
 
### Applicable Issues

https://github.com/cryptape/cita-common/issues/267


kaikai1024:
#604

PR #614, opened 2019-05-30
Fix circle ci

kaikai1024:
### Description of the Change

Make every test independent.

### Applicable Issues

#614

### DIFF
--- a/tests/integrate_test/cita_tls_basic.sh
+++ b/tests/integrate_test/cita_tls_basic.sh
@@ -16,12 +16,7 @@ cd ${BINARY_DIR}
 echo "DONE"
 
 ################################################################################
-echo -n "1) cleanup   ...  "
-cleanup
-echo "DONE"
-
-################################################################################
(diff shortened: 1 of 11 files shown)


tests/integrate_test/cita_tls_basic.sh
```diff
@@ -16,12 +16,7 @@ cd ${BINARY_DIR}
 echo "DONE"
 
 ################################################################################
-echo -n "1) cleanup   ...  "
```
ouwenkg: Why remove these cleanup?

kaikai1024: It's useless since every test is an independent environment. 

Why did we need it before? because all tests run on one environment.

the snapshot test does not always work when run after other tests, maybe the cleanup is not really `cleanup`

ouwenkg: OK

PR #629, opened 2019-06-10
fix: fix cita.sh for NODE_PATH should not consider CITA_BIN path

### LINKED ISSUE #626: Why data dir is binding to CITA bin dir?
```
# which cita
/opt/cita-run/cita_secp256k1_sha3/bin/cita
```

```
# pwd
/data
root@cita_node:/data# ls -d1 /data/test-chain/0/
/data/test-chain/0/

root@cita_node:/data# cita bebop setup test-chain/0
No such node directory: test-chain/0
```

this behave was all right in CITA v0.20.x (using `cita setup test-chain/0`).

account to the source code:
https://github.com/cryptape/cita/blob/develop/scripts/cita.sh#L35

I found that

```
NODE_PATH="$(dirname "$CITA_BIN")/${NODE_NAME}"
```

https://github.com/cryptape/cita/blob/develop/scripts/cita.sh#L400:L403

```
# Make sure the node directory exists
if [ ! -d "${NODE_PATH}" ]; then
    echo "No such node directory: ${NODE_NAME}"
    exit 1

```

Why data dir is MUST binding to bin dir? 
This is quite incomprehensible since bin/data/config/log dir should be able to be separated then we can update the bin without touching the data,config or log dir.

here is a quick fix:

```
NODE_PATH="$(realpath ${NODE_NAME})"
```

rainchen:
fix #626


scripts/cita.sh
```diff
@@ -32,7 +32,7 @@ SERVICES=( forever auth bft chain executor jsonrpc network )
 SCRIPT=$(basename "$0")
 COMMAND=$1
 NODE_NAME=$2
-NODE_PATH="$(dirname "$CITA_BIN")/${NODE_NAME}"
+NODE_PATH="$(realpath ${NODE_NAME})"
```
kaikai1024:
```suggestion
NODE_PATH=$(realpath "${NODE_NAME}")
```

And check the https://github.com/cryptape/cita/blob/develop/scripts/cita.sh#L6-L10

rainchen: updated

kaikai1024: https://unix.stackexchange.com/questions/136494/whats-the-difference-between-realpath-and-readlink-f

PR #644, opened 2019-06-12
Fix: should not use error message in default auto exec.

leeyr338:
Should not use error message in default auto exec.

The function `conf.auto_exec` will run each block, the `Error`  message will confuse to our users.

```
2019-06-11T12:57:07.394285300+00:00 - ERROR - Use the default autoEXEC.
2019-06-11T12:57:08.926864406+00:00 - INFO - executor init, current_height: 3724098, current_hash: 0x8d5abdbaab8469a1af27402e2f44026da6a21dac5f98131e475a4b6cd4df0ea0
2019-06-11T12:57:08.934952743+00:00 - INFO - Use default quota price
2019-06-11T12:57:08.948182035+00:00 - INFO - Use default auto exec quota limit.
2019-06-11T12:57:08.970409072+00:00 - ERROR - Use the default autoEXEC.
2019-06-11T12:57:09.749927095+00:00 - INFO - Use default quota price
2019-06-11T12:57:09.757813566+00:00 - INFO - Use default auto exec quota limit.
2019-06-11T12:57:09.773810266+00:00 - ERROR - Use the default autoEXEC.
```

### DIFF
--- a/cita-executor/core/src/contracts/solc/sys_config.rs
+++ b/cita-executor/core/src/contracts/solc/sys_config.rs
@@ -341,7 +341,7 @@ impl<'a> SysConfig<'a> {
     }
 
     pub fn default_auto_exec() -> bool {
-        error!("Use the default autoEXEC.");
+        info!("Use the default autoEXEC.");
         false
     }
 }


kaikai1024:
this is a missed modification.
check the https://github.com/cryptape/cita/blob/develop/cita-executor/core/src/contracts/solc/node_manager.rs#L110

maybe it's better to use `info` as same.

and check the other `default_*` part of this file. maybe use:

`should not use error message in sys_config` or something like this.

cita-executor/core/src/contracts/solc/sys_config.rs
```diff
@@ -341,7 +341,7 @@ impl<'a> SysConfig<'a> {
     }
 
     pub fn default_auto_exec() -> bool {
-        error!("Use the default autoEXEC.");
+        debug!("Use the default autoEXEC.");
```
ouwenkg: `Use the default auto exec config` ?

leeyr338:
> this is a missed modification.
> check the https://github.com/cryptape/cita/blob/develop/cita-executor/core/src/contracts/solc/node_manager.rs#L110
> 
> maybe it's better to use `info` as same.
> 
> and check the other `default_*` part of this file. maybe use:
> 
> `should not use error message in sys_config` or something like this.

Yes! `info` is okay!

cita-executor/core/src/contracts/solc/sys_config.rs
leeyr338: Some users upgrade CITA from v0.19 t0 v0.23, and there is no `getAutoExec()` in  genesis. So it uses default config.

Issue #492, opened 2019-04-22
The difference of getVersion of `RPC` and `VersionManger`


kaikai1024:
## Description

There are two `getVersion`. It' confused for users.

* The rpc to get the version of software
* The interfae of `VersionManager` system contract to get the version of protocol

Should be different.

## Additional Information

https://docs.citahub.com/zh-CN/next/cita/rpc-guide/rpc#getversion

https://docs.citahub.com/zh-CN/next/cita/system/version

kaikai1024:
@leeyr338 @ouwenkg

kaikai1024:
The doc should be very clear.

kaikai1024:
Rename: `getSoftwareVersion`

kaikai1024:
What's next?

True deprecated.

A description about upgrade:

* amend version manager contract code
* amend the key

cita-executor uses `getProtocolVersion`.

PR #651, opened 2019-06-18
Protocol version

kaikai1024:
### Description of the Change

Use new interface:

* getProtocolVersion
* setProtocolVersion

old interfaces still in use.

> @notice Deprecated.

### Applicable Issues

#492

### DIFF
--- a/cita-executor/core/src/contracts/solc/version_management.rs
+++ b/cita-executor/core/src/contracts/solc/version_management.rs
@@ -1,5 +1,5 @@
 // CITA
-// Copyright 2016-2018 Cryptape Technologies LLC.
+// Copyright 2016-2019 Cryptape Technologies LLC.
 
 // This program is free software: you can redistribute it
 // and/or modify it under the terms of the GNU General Public
@@ -30,6 +30,8 @@ use types::ids::BlockId;
 use types::reserved_addresses;
 
 lazy_static! {
+    // Will use `getProtocolVersion` at next version after `v0.25.0`.
+    // And the `getVersion` will be *Deprecated*.
     static ref VERSION_HASH: Vec<u8> = method_tools::encode_to_vec(b"getVersion()");
     static ref CONTRACT_ADDRESS: Address =
         Address::from_str(reserved_addresses::VERSION_MANAGEMENT).unwrap();
--- a/scripts/contracts/contracts.yml
+++ b/scripts/contracts/contracts.yml
@@ -221,5 +221,7 @@ PermissionContracts:
       address: '0xffffffffffffffffffffffffffffffffff021028'
       contracts:
         - VersionManager
+        - VersionManager
       functions:
         - 'setVersion(uint32)'
+        - 'setProtocolVersion(uint32)'
--- a/scripts/contracts/src/system/VersionManager.sol
+++ b/scripts/contracts/src/system/VersionManager.sol
@@ -23,7 +23,7 @@ contract VersionManager is IVersionManager, ReservedAddrPublic {
         version = _version;
     }
 
-    function setVersion(uint32 _version)
+    function setProtocolVersion(uint32 _version)
         public
         onlyAdmin
     {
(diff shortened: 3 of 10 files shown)


ouwenkg:
ci...

kaikai1024:
> ci...

ok. Need to update the genesis.

kaikai1024:
There's a problem: It's required to amend the code of version manager when the chain needs a upgrade to the next version.
I'll check if i can skip the upgrade

kaikai1024:
One version to prepare the upgrade.
And will update the doc to show the info.

ouwenkg:
> There's a problem: It's required to amend the code of version manager when the chain needs a upgrade to the next version.
> I'll check if i can skip the upgrade

We need a script tool or operation guide docs.

kaikai1024:
> We need a script tool or operation guide docs.

yes. but i want the upgrade is a choice for users. maybe a guide doc or use `cita-cli`, i don't want to depend the txtool like `amend_system_contracts`. 
by the way, `amend_system_contracts` is useless.

scripts/contracts/src/system/VersionManager.sol
```diff
@@ -37,11 +37,28 @@ contract VersionManager is IVersionManager, ReservedAddrPublic {
...
 
-    function getVersion()
+    /// @notice Deprecated. Check the setProtocolVersion
+    function setVersion(uint32 _version)
+        public
+        onlyAdmin
+    {
+        setProtocolVersion(_version);
+    }
+
```
ouwenkg: 这样实现不太好吧, 保持原来的 `set_version` 不动, 然后新增加一个 `setProtocolVersion` 接口就好了.

kaikai1024: Don't get your point. 
It's just like what you say, old version uses the old interface, new version uses the new interface.

ouwenkg: OK, i understand.

ouwenkg:
Please update `contracts.yml` permission section.

kaikai1024:
> Please update `contracts.yml` permission section.

opp, I forget this.

Change the system contracts is a bad idea....
Actually it should change the rpc interface(getSoftwareVersion), but i don't want to trouble sdk or test.

ouwenkg:
> > Please update `contracts.yml` permission section.
> 
> opp, I forget this.
> 
> Change the system contracts is a bad idea....
> Actually it should change the rpc interface(getSoftwareVersion), but i don't want to trouble sdk or test.


> > Please update `contracts.yml` permission section.
> 
> opp, I forget this.
> 
> Change the system contracts is a bad idea....
> Actually it should change the rpc interface(getSoftwareVersion), but i don't want to trouble sdk or test.

No matter which way we choose, trouble sdk or test is inevitable?

kaikai1024:
> No matter which way we choose, trouble sdk or test is inevitable?

Not a level. 
one is a base rpc interface(both sdk and test need to work to fix it.), the other one is just a contract interface call(sdk can skip it and test can add a new case.).

cita-executor/core/src/contracts/solc/version_management.rs
```diff
@@ -30,6 +30,8 @@ use types::ids::BlockId;
 use types::reserved_addresses;
 
 lazy_static! {
+    // Will use `getProtocolVersion` at next version after `v0.25.0`.
+    // And the `getVersion` will be *Deprecated*.
     static ref VERSION_HASH: Vec<u8> = method_tools::encode_to_vec(b"getVersion()");
```
ouwenkg: Why next version?

scripts/contracts/contracts.yml
```diff
@@ -223,3 +223,4 @@ PermissionContracts:
         - VersionManager
       functions:
         - 'setVersion(uint32)'
+        - 'setProtocolVersion(uint32)'
```
ouwenkg: Please update permission section in genesis module.

cita-executor/core/src/contracts/solc/version_management.rs
kaikai1024: The new coming version will not enforce user to amend code when upgrade.

scripts/contracts/contracts.yml
kaikai1024: ok

scripts/contracts/contracts.yml
```diff
@@ -223,3 +223,4 @@ PermissionContracts:
         - VersionManager
       functions:
```
ouwenkg: 这里应该有两个 VersionManager 吧

cita-executor/core/src/contracts/solc/version_management.rs
ouwenkg: Ok

scripts/contracts/contracts.yml
kaikai1024: yes

Issue #650, opened 2019-06-14
add test about permission deny


ouwenkg:
## Add test about permission deny

[Description of the issue]

```rs
 #[test]
    fn test_contract_address_from_permission_denied() {
        let keypair = KeyPair::gen_keypair();
        let privkey = keypair.privkey();

        let mut executor =
            helpers::init_executor(vec![("SysConfig.checkCreateContractPermission", "true")]);

        let chain = helpers::init_chain();
        let data = helpers::generate_contract();
        let block = helpers::create_block(&executor, Address::from(0), &data, (0, 1), &privkey);
        let inchain = chain.clone();
        let txs = block.body().transactions().clone();
        let hash = txs[0].hash();
        let h = executor.get_current_height() + 1;

        let closed_block = executor.into_fsm(block.clone());
        let executed_result = executor.grow(closed_block);
        inchain.set_block_body(h, &block);
        inchain.set_db_result(&executed_result, &block);

        let receipt = chain
            .localized_receipt(hash)
            .expect("failed to get localized_receipt");
        assert_eq!(receipt.contract_address, None);
        assert_eq!(receipt.error, Some(ReceiptError::NoContractPermission));
    }
```
Add this test case in ci check.

ouwenkg:
#638

PR #669, opened 2019-06-26
add permission denied test

ouwenkg:
### Description of the Change

1. Add permission denied test
2. Put common test tools together into `txtool_utils.py`, and make py-tests more clear. 
3. Seperate features relevant test from original `cita_charge_mode.py`, only transfer cases remained
now.

### Applicable Issues

#650

### DIFF
--- a/tests/integrate_test/cita_charge_mode.sh
+++ b/tests/integrate_test/cita_charge_mode.sh
@@ -47,10 +47,8 @@ main() {
     echo -n "1) generate config  ...  "
     create_config \
         --contract_arguments "SysConfig.checkFeeBackPlatform=true" \
-        --contract_arguments "SysConfig.checkCreateContractPermission=true" \
         --contract_arguments "SysConfig.economicalModel=1" \
         --contract_arguments "VersionManager.version=0" \
-        --contract_arguments "PriceManager.quotaPrice=1000000" \
         --contract_arguments "SysConfig.chainOwner=0x36a60d575b0dee0423abb6a57dbc6ca60bf47545"
     echo "DONE"
 
--- a/tests/integrate_test/cita_charge_mode.sh
+++ b/tests/integrate_test/cita_charge_mode.sh
@@ -37,6 +37,13 @@ test_fee_back() {
     cd ../../..
 }
 
+test_perm_denied() {
+    local version=$1
+    cd ./scripts/txtool/txtool
+    python3 "${SOURCE_DIR}"/tests/integrate_test/test_perm_denied.py --version="$version"
+    cd ../../..
+}
+
 update_version() {
     local version=$1
     cd ./scripts/txtool/txtool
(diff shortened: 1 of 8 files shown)


tests/integrate_test/cita_charge_mode.sh
```diff
@@ -54,6 +47,7 @@ main() {
     echo -n "1) generate config  ...  "
     create_config \
         --contract_arguments "SysConfig.checkFeeBackPlatform=true" \
+        --contract_arguments "SysConfig.checkCreateContractPermission=true" \
```
kaikai1024: No need.
and `--contract_arguments "SysConfig.checkFeeBackPlatform=true"`

tests/integrate_test/cita_features_test.sh
```diff
@@ -0,0 +1,62 @@
...
+    . ${SOURCE_DIR}/tests/integrate_test/util.sh
+    cd "${BINARY_DIR}"
+    echo "DONE"
+
+    echo -n "1) generate config  ...  "
+    create_config \
+        --contract_arguments "SysConfig.checkFeeBackPlatform=true" \
+        --contract_arguments "SysConfig.checkCreateContractPermission=true" \
+        --contract_arguments "SysConfig.economicalModel=1" \
+        --contract_arguments "VersionManager.version=0" \
```
kaikai1024: What about version 1 and 2?
Why do not use version 2 directly?

tests/integrate_test/cita_charge_mode.sh
ouwenkg: permission setting is needless, checkFeeBackPlatform is used while checking transferd balance.

tests/integrate_test/cita_features_test.sh
ouwenkg: I just copy from `cita_charge_mode`, version=2 is ok.

tests/integrate_test/cita_charge_mode.sh
kaikai1024: Is not checkFeeBackFlatform used for fee_back test?

tests/integrate_test/cita_features_test.sh
kaikai1024: ok.
`cita_charge_mode` testes both version and version 1. Your copy seems miss version 1.

tests/integrate_test/cita_charge_mode.sh
ouwenkg: Setting `SysConfig.checkFeeBackPlatform=true`, the transaction fee goes to chain owner,
when remove this config, tx fees goes to miner which causes original transfer test cases fail.

I do not want to change orginal test logic, so remain checkFeeBackPlatform=true.

kaikai1024: I see.

tests/integrate_test/cita_features_test.sh
ouwenkg: I remove version config at start, just use latest version 2

PR #676, opened 2019-07-02
Refactor cita-bft 

rev-chaos:
## Description

The proof in bft-rs is already compatible with cita. This PR is used for ci test.


kaikai1024:
:+1:  
You can create a PR for cita-bft now

jerry-yu:
old

PR #679, opened 2019-07-04
fix bug for index greater than array's limit

jerry-yu:
fix bug for index greater than array's limit

### DIFF
--- a/cita-executor/core/src/executive.rs
+++ b/cita-executor/core/src/executive.rs
@@ -232,6 +232,9 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
     }
 
     fn transact_set_abi(&mut self, data: &[u8]) -> bool {
+        if data.len() <= 20 {
+            return false;
+        }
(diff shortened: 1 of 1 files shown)


cita-executor/core/src/executive.rs
```diff
@@ -232,6 +232,9 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
     }
 
     fn transact_set_abi(&mut self, data: &[u8]) -> bool {
+        if data.len() < 20 {
```
kaikai1024: What if `data.len() == 20`?

ouwenkg:
:+1: 

```rs
fn transact_get_kv_h256(&mut self, data: &[u8]) -> Option<H256> {
        let account = H160::from(&data[0..20]);
        let key = H256::from_slice(&data[20..52]);
        self.state.storage_at(&account, &key).ok()
}
```

Should we check length in get function?

kaikai1024:
Better to add unit test about the bug.

PR #680, opened 2019-07-04
Fix the security audit: upgrade smallvec.

kaikai1024:
# Description of the Change

FIx the check:

```
make security_audit
```


kaikai1024:
#679

PR #682, opened 2019-07-09
Split account_entry from state mod.

kaikai1024:
### Description of the Change

Split account_entry from state mod.

*`remove-gpl` branch needs to rebase it.*

### DIFF
--- a/cita-executor/core/src/state/account_entry.rs
+++ b/cita-executor/core/src/state/account_entry.rs
@@ -1,0 +1,120 @@
+// Copyright 2015-2017 Parity Technologies (UK) Ltd.
+// This file is part of Parity.
+
+// This software is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This software is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


cita-executor/core/src/state/mod.rs
```diff
@@ -61,106 +63,6 @@ pub struct ApplyOutcome {
...
-            }
-            None => self.account = None,
-        }
-    }
-
-    pub fn account(&self) -> Option<&Account> {
-        self.account.as_ref()
-    }
-}
-
```
ouwenkg: Defined again!

cita-executor/core/src/state/account_entry.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+}
+
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
+pub struct StateProof {
+    address: Address,
+    account_proof: Vec<Bytes>,
+    key: H256,
+    value_proof: Vec<Bytes>,
+}
```
ouwenkg: The `StateProof` struct duplicated defined here.

kaikai1024: yes, it's a mistake

cita-executor/core/src/state/mod.rs
kaikai1024: done

PR #683, opened 2019-07-11
Remove the super admin of state.

kaikai1024:
### Description of the Change

super admin account is no need for state.

### Alternate Design

NO

### Benefits

Keep the state clean.

### Possible Drawbacks

call the amend address using solidity assembly.

### Verification Process

Unit test

### Applicable Issues

None

### DIFF
--- a/cita-executor/core/src/tests/amend_data_test.rs
+++ b/cita-executor/core/src/tests/amend_data_test.rs
@@ -1,5 +1,5 @@
 // CITA
-// Copyright 2016-2018 Cryptape Technologies LLC.
+// Copyright 2016-2019 Cryptape Technologies LLC.
 
 // This program is free software: you can redistribute it
 // and/or modify it under the terms of the GNU General Public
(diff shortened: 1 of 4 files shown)


cita-executor/core/src/tests/amend_data_test.rs
```diff
@@ -66,53 +108,38 @@ fn call_vm(
...
     data.append(&mut key.to_vec());
+    // Get value from key use transact interface
+    // `value=4` means the operation of getting value from key.
+    let result = transact(&mut state, &data, U256::from(4), true);
+    assert!(result.is_ok());
+    // TODO Add output of transact in executive.rs when amend get-value
+    // println!("result: {:?}", result.clone().unwrap());
+
+    // Get value from key use call interface
+    let address: Address = reserved_addresses::AMEND_ADDRESS.into();
```
ouwenkg: It's better to slipt this test into two cases `superAdmin` or `Not superAdmin`

kaikai1024: It just fix the older test. so if follow the old style.

There are othere problems beside what you say:

* the other amend operation
* the output(check the todo), maybe it's the problem of progress of executive.

kaikai1024:
Already check the super admin:

https://github.com/cryptape/cita/blob/develop/cita-executor/core/src/executive.rs#L342-L350

ouwenkg:
Please update the branch you know.

kaikai1024:
DONE

Issue #677, opened 2019-07-04
cost same quota when send a transaction with data or not, even the data is very big 


luqz:
## Description

cost same quota when send a transaction with data or not, even the data is very big 
## Steps to Reproduce

```
sdk.base.getBlockNumber()
    .then(async current => {
      let testString = "测试"
      for (let index = 0; index < 20; index++) {
        testString += testString
      }

      let blcokNumber = await sdk.base.getBlockNumber()
      let metaData = await sdk.base.getMetaData()

      let tx = {
        to: "0x1072fd65138cf36cd5d3e77661c1138e475ebf4d",
        value: "0x1",
        data: sdk.utils.stringToHex(testString),
        quota: 21000,
        chainId: metaData.chainIdV1,
        version: metaData.version,
        validUntilBlock: blcokNumber + 99,
        privateKey: "0xe37861d07588985bf01d5e4b314477d9d11f0189e446d009930e5b20be3d19f1"
      }

      let res = await sdk.base.sendTransaction(tx)
      console.log(res.hash)
      let receipt = await sdk.listeners.listenToTransactionReceiptEx(res.hash)
      console.log(receipt)

    })
```
**Expected behavior**: quotaUsed is different by tx.data.length 

**Actual behavior**: quotaUsed always 21000

**Reproduce how often**: every time

## Versions

v0.25.0

## Additional Information

I think this behavior need be change soon.

kaikai1024:
21000 is the base gas of a tx.
Is the `0x1072fd65138cf36cd5d3e77661c1138e475ebf4d` a contract address? and the what's the data?

luqz:
> 21000 is the base gas of a tx.
> Is the `0x1072fd65138cf36cd5d3e77661c1138e475ebf4d` a contract address? and the what's the data?
It's not a contract address. The data is just for test, it‘s meaningless, maybe something to record。

kaikai1024:
> It's not a contract address.

Then it's normal for CITA for now. It's a tx with the data that will not be executed. so it will cost the base gas no matter how long the data is. 

But i thinkg it's not a normal situation, either.

I remember there's an EIP about this. wait a minute

kaikai1024:
I will add some descriptions about this later.

ouwenkg:
> ~I will add some descriptions about this later.~
> 
> CITA _executes_ txs after consensus, while ethereum executes txs _before_ consensus.
> This tx can not be executed successful on ethereum, so it will not be included in block.
[...]

PR #685, opened 2019-07-11
increase gas by charging for transaction data

luqz:
### Description of the Change
1. increase gas by charging for transaction data;
2. add verify of min quota needed in auth.

### DIFF
--- a/cita-auth/src/block_verify.rs
+++ b/cita-auth/src/block_verify.rs
@@ -15,6 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use crate::handler::verify_base_quota_required;
 use cita_types::traits::LowerHex;
 use cita_types::Address;
 use crypto::{pubkey_to_address, PubKey};
@@ -49,6 +50,10 @@ impl<'a> BlockVerify<'a> {
             let signer = pubkey_to_address(&PubKey::from(tx.get_signer()));
 
             if block_quota_limit < quota {
+                return false;
+            }
+
+            if !verify_base_quota_required(tx.get_transaction_with_sig().get_transaction()) {
                 return false;
             }
 
--- a/cita-auth/src/handler.rs
+++ b/cita-auth/src/handler.rs
@@ -24,9 +24,10 @@ use cita_types::traits::LowerHex;
 use cita_types::{clean_0x, Address, H256, U256};
 use crypto::{pubkey_to_address, PubKey, Sign, Signature, SIGNATURE_BYTES_LEN};
 use error::ErrorCode;
+use evm::Schedule;
 use jsonrpc_types::rpc_types::TxResponse;
 use libproto::auth::{Miscellaneous, MiscellaneousReq};
-use libproto::blockchain::{AccountGasLimit, SignedTransaction};
+use libproto::blockchain::{AccountGasLimit, SignedTransaction, Transaction};
 use libproto::router::{MsgType, RoutingKey, SubModules};
 use libproto::snapshot::{Cmd, Resp, SnapshotReq, SnapshotResp};
 use libproto::{
@@ -183,7 +184,7 @@ impl MsgHandler {
         true
     }
 
-    // verify to and version
+    // verify to and version and min quota
     fn verify_request(&self, req: &Request) -> Result<(), Error> {
         let un_tx = req.get_un_tx();
         let tx = un_tx.get_transaction();
@@ -213,6 +214,10 @@ impl MsgHandler {
         } else {
             error!("unexpected version {}!", tx_version);
             return Err(Error::InvalidValue);
+        }
+
+        if !verify_base_quota_required(tx) {
+            return Err(Error::QuotaNotEnough);
         }
 
         Ok(())
@@ -1089,4 +1094,22 @@ fn snapshot_response(sender: &Sender<(String, Vec<u8>)>, ack: Resp, flag: bool) 
             (&msg).try_into().unwrap(),
         ))
         .unwrap();
+}
+
+// only verify if tx.version > 2
+pub fn verify_base_quota_required(tx: &Transaction) -> bool {
+    match tx.get_version() {
+        0...2 => true,
+        _ => {
+            let schedule = Schedule::new_v1();
+            let to = tx.get_to_v1();
+            if to.is_empty() || Address::from(to) == Address::zero() {
+                tx.get_quota() as usize
+                    >= tx.data.len() * schedule.tx_data_non_zero_gas + schedule.tx_create_gas
+            } else {
+                tx.get_quota() as usize
+                    >= tx.data.len() * schedule.tx_data_non_zero_gas + schedule.tx_gas
+            }
+        }
+    }
 }
--- a/cita-executor/core/src/executive.rs
+++ b/cita-executor/core/src/executive.rs
@@ -330,6 +330,9 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
             Action::Create => schedule.tx_create_gas,
             Action::GoCreate => schedule.tx_create_gas,
             _ => schedule.tx_gas,
+        } + match t.version {
+            0...2 => 0,
+            _ => t.data.len() * schedule.tx_data_non_zero_gas,
         };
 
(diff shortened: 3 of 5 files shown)


cita-auth/src/block_verify.rs
```diff
@@ -71,6 +75,23 @@ impl<'a> BlockVerify<'a> {
         true
     }
 
+    fn verify_min_quota(&self, sig_tx: &SignedTransaction) -> bool {
```
rainchen: should add a comment for this method

cita-executor/core/src/executive.rs
```diff
@@ -326,11 +326,16 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
+            0...2 => 0,
+            _ => t.data.len() * 4,
+        };
+
         let base_gas_required = match t.action {
             Action::Create => schedule.tx_create_gas,
             Action::GoCreate => schedule.tx_create_gas,
             _ => schedule.tx_gas,
-        };
+        } + gas_used_by_data_length;
```
rainchen: I think it's better to add a new variable named like `data_gas_required = base_gas_required  + gas_used_by_data_length`, but not channging the original `base_gas_required` 's definition

rainchen:
@luqz  `Check Clippy - Failed` please update & rebase

rainchen:
@luqz you should use `git amend` to mange your commits but should not add a new commit to just add a line of comment

cita-auth/src/block_verify.rs
```diff
@@ -71,6 +75,26 @@ impl<'a> BlockVerify<'a> {
         true
     }
 
+    // min quota is 21000 or (32000 if tx.to == 0) + (4 per byte)
+    // tx quota: 21000
+    // create contract: 32000
+    fn verify_min_quota(&self, sig_tx: &SignedTransaction) -> bool {
+        let tx = sig_tx.get_transaction_with_sig().get_transaction();
+        let tx_version = tx.get_version();
+        if tx_version < 3 {
```
kaikai1024: It always return true. how to verify min quota?

cita-auth/src/block_verify.rs
kaikai1024: If you follow the way of Ethereum, then:
```
21000 + (68 or 4 per byte) + (32000 if tx.to == 0)
```

create_contract: 53000

cita-auth/src/handler.rs
```diff
@@ -183,7 +183,27 @@ impl MsgHandler {
         true
     }
 
-    // verify to and version
+    // min quota is 21000 or (32000 if tx.to == 0) + (4 per byte)
+    // tx quota: 21000
+    // create contract: 32000
+    fn verify_min_quota(&self, req: &Request) -> bool {
```
kaikai1024: Why define two same function?

cita-executor/core/src/executive.rs
```diff
@@ -326,11 +326,16 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
 
         let schedule = Schedule::new_v1();
 
+        let gas_used_by_data_length = match t.version {
```
kaikai1024: nice

cita-auth/src/block_verify.rs
luqz: I will change amend it.

cita-auth/src/block_verify.rs
luqz: only verify tx with new tx_version

kaikai1024: Get it.
Actually, it's no need to consider the version at `cita-auth` module

cita-auth/src/handler.rs
```diff
@@ -1090,3 +1094,24 @@ fn snapshot_response(sender: &Sender<(String, Vec<u8>)>, ack: Resp, flag: bool)
         ))
         .unwrap();
 }
+
+// min quota is 21000 (or 53000 if tx.to == 0) + (4 per byte)
+// tx quota: 21000
+// create contract: 53000
+// only verify if tx.version > 2
+pub fn verify_min_quota(tx: &Transaction) -> bool {
+    match tx.get_version() {
```
kaikai1024: Actually, it's no need to consider the version at cita-auth module

cita-auth/src/block_verify.rs
luqz: I think it's needed for tx in early blocks.

cita-auth/src/handler.rs
luqz: This function is used in block verify too. In early block, it's need to consider the version.

kaikai1024: I see.

cita-auth/src/handler.rs
```diff
@@ -1090,3 +1094,24 @@ fn snapshot_response(sender: &Sender<(String, Vec<u8>)>, ack: Resp, flag: bool)
...
+        0...2 => true,
+        _ => {
+            let quota = tx.get_quota();
+            let to = tx.get_to_v1();
+            if to.is_empty() || Address::from(to) == Address::zero() {
+                let min_quota = 53000 + 4 * tx.data.len();
```
leeyr338: ref: https://github.com/cryptape/cita/blob/develop/cita-executor/evm/src/schedule.rs#L100
The quota for create a contract, is
```
tx_create_gas: 53_000
```
but not `32000`.
`32000` is the quota for `CREATE` instruction.

BTW, don't use the number like `21000` and `32000` directly, it is better to define a Macro or use `let schedule = Schedule::new_v1();` for it.

cita-executor/core/src/executive.rs
leeyr338: This should be `+ gas_used_by_data_length * schedule.create_data_gas;` ?

kaikai1024: Why `* schedule.create_data_gas`?

leeyr338: `store_gas_used = U256::from(t.data.len() * schedule.create_data_gas);`

leeyr338: `store_gas_used = U256::from(t.data.len() * schedule.create_data_gas);`

kaikai1024: It's base gas instead of the store gas.  stored gas used will be calculated in vm.

> // gas_prepare = 21000 + (68 or 4 per byte) + (32000 if tx.to == 0)

luqz: I think t.data.len() * schedule.create_data_gas is ok. 
Now, schedule.create_data_gas is 200, default BQL is 0x4000000, 0x4000000 / 200 = 5368709.12.

cita-auth/src/handler.rs
luqz: Struct Schedule is in the evm crate, maybe it's no worth to use it here.
I don't know how to define a Macro in this case.

cita-executor/core/src/executive.rs
kaikai1024: The difference:
* one is before vm, and one is in vm
* one is tx data, and one will be stored in storage

What i want to say is that the base gas is not related to the instruction of vm.

Of course, We can define our rule about this or follow the Ethereum.

leeyr338: But we also use `200` for each byte of our "存证" data. Should we follow this rule?

cita-executor/core/src/executive.rs
```diff
@@ -326,11 +326,16 @@ impl<'a, B: 'a + StateBackend> Executive<'a, B> {
...
+            0...2 => 0,
+            _ => t.data.len() * schedule.create_data_gas,
```
leeyr338: This value is `gas_used_by_data_length`, can't represent its actual meaning. Pls change the value name.

cita-executor/core/src/executive.rs
kaikai1024: In my humble opinion, stored data is actually tx data, it should cost base gas(ignore the already implement): 
```
21000 + (68 per byte)
```

leeyr338: Okay! I agree with you!

luqz: I have changed it to 21000 + (68 per byte)

kaikai1024:
#677

PR #687, opened 2019-07-12
[Feature] remove EVM: 1. add cita-vm. [skip ci]

leeyr338:
1. add cita-vm.

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -1,0 +1,27 @@
+use cita_trie::DB;
+use cita_vm::{state::State, BlockDataProvider, Config};
+use std::cell::RefCell;
+use std::sync::Arc;
+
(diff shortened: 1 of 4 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -0,0 +1,30 @@
+// CITA
```
kaikai1024: No need to add this.
Change the GPL part if you want to.

ouwenkg: All these license docs could be removed after our task.

kaikai1024: Then, why do we add it first?

ouwenkg: Adding now or Removing together later is accecpted to me.

leeyr338: okay, delete it.

PR #692, opened 2019-07-16
[Feature] remove EVM: add tx_gas_schedule and more logic for CitaExecutive:exec

leeyr338:
- add tx_gas_schedule and more logic for CitaExecutive:exec.
- add struct `ExecParams`, that some one may use.

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -1,4 +1,5 @@
 use cita_trie::DB;
+use cita_types::{Address, H160, H256, U256, U512};
 use cita_vm::{
     state::{Error as StateError, State, StateObjectInfo},
     BlockDataProvider, Config, Error as VMError,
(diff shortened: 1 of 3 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -113,3 +357,34 @@ impl From<AuthenticationError> for ExecutionError {
...
+    /// Returns default ActionParams initialized with zeros
+    fn default() -> ExecParams {
+        ExecParams {
+            code_address: Address::new(),
+            sender: Address::new(),
+            gas: U256::zero(),
+            data: None,
+        }
+    }
+}
```
ouwenkg: Could we put struct `ExecParams`, `TxExecResult` out in a separated `.rs`, and make cita_executive more clearly and simple ?  FYI

leeyr338: I will refine it later.

PR #694, opened 2019-07-17
[Feature] remove EVM: add create contract logic for CitaExecutive:exec

leeyr338:
[Feature] remove EVM: add create contract logic for CitaExecutive:exec.

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -1,8 +1,9 @@
 use cita_trie::DB;
 use cita_types::{Address, H160, H256, U256, U512};
 use cita_vm::{
+    evm::{Context as EVMContext, InterpreterResult, Log as EVMLog},
     state::{Error as StateError, State, StateObjectInfo},
-    BlockDataProvider, Config, Error as VMError,
+    BlockDataProvider, Config as VMConfig, Error as VMError, Transaction as EVMTransaction,
 };
 use std::cell::RefCell;
 use std::error::Error;
@@ -11,11 +12,16 @@ use std::sync::Arc;
 use util::Bytes;
 
 use crate::authentication::{check_permission, AuthenticationError};
-use crate::core_types::{receipt::Receipt, TypesError};
+use crate::core_types::{Bloom, BloomInput, Hash, LogEntry, Receipt, TypesError};
 use crate::libexecutor::economical_model::EconomicalModel;
 use crate::libexecutor::sys_config::BlockSysConfig;
 use crate::tx_gas_schedule::TxGasSchedule;
 use crate::types::transaction::{Action, SignedTransaction};
+use crate::types::BlockNumber;
+
+/// Simple vector of hashes, should be at most 256 items large, can be smaller if being used
+/// for a block whose number is less than 257.
+pub type LastHashes = Vec<H256>;
 
 ///amend the abi data
 const AMEND_ABI: u32 = 1;
@@ -28,9 +34,9 @@ const AMEND_ACCOUNT_BALANCE: u32 = 5;
 
 // FIXME: CITAExecutive need rename to Executive after all works ready.
 pub struct CitaExecutive<B> {
-    _block_provider: Arc<BlockDataProvider>,
+    block_provider: Arc<BlockDataProvider>,
     state_provider: Arc<RefCell<State<B>>>,
-    _config: Config,
+    env_info: EnvInfo,
     economical_model: EconomicalModel,
 }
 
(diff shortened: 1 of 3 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -123,50 +130,42 @@ impl<B: DB + 'static> CitaExecutive<B> {
...
                 // But for the data compatible, just let it as tx_gas_schedule.create_data_gas for now.
                 let store_gas_used = U256::from(t.data.len() * tx_gas_schedule.create_data_gas);
-                if let Some(gas_left) = init_gas.checked_sub(store_gas_used) {
-                    Ok(TxExecResult { gas_left })
-                } else {
-                    return Err(ExecutionError::VM(VMError::NotEnoughBaseGas));
-                }
+                let mut receipt = Receipt::default();
+                receipt.quota_used = store_gas_used.as_u64();
+                Ok(receipt)
```
ouwenkg: How to handle error situation, such as `init_gas < store_gas_used`?

cita-executor/core/src/cita_executive.rs
```diff
@@ -256,50 +255,165 @@ impl<B: DB + 'static> CitaExecutive<B> {
         true
     }
 
-    fn call_amend_data(
-        &mut self,
-        init_gas: U256,
-        value: U256,
-        data: Option<Bytes>,
-    ) -> Result<TxExecResult, ExecutionError> {
+    fn call_amend_data(&mut self, value: U256, data: Option<Bytes>) -> Result<(), ExecutionError> {
```
ouwenkg: No need to return `TxExecResult` ?

cita-executor/core/src/cita_executive.rs
```diff
@@ -123,50 +130,42 @@ impl<B: DB + 'static> CitaExecutive<B> {
...
+                    code_address: None,
+                    sender,
+                    to_address: None,
+                    gas: init_gas,
+                    gas_price: t.gas_price(),
+                    value: t.value,
+                    nonce: U256::zero(),
+                    data: t.data.clone(),
+                };
+
```
ouwenkg: We should get params by re-translating transaction info.

cita-executor/core/src/cita_executive.rs
leeyr338: I need some logic to handle the result, but haven't finished yet, I plan to handle this situation there.
But maybe it is better to handle this situation here. I will change this logic.

cita-executor/core/src/cita_executive.rs
leeyr338: Actually, we have done this. And I just need only this params to create a Contract.

cita-executor/core/src/cita_executive.rs
leeyr338: Yes, We do not this the result `gas_left`,  because it do not any `gas` in `amend` logic.

cita-executor/core/src/cita_executive.rs
```diff
@@ -256,50 +259,165 @@ impl<B: DB + 'static> CitaExecutive<B> {
...
+fn build_evm_context(env_info: &EnvInfo) -> EVMContext {
+    EVMContext {
+        gas_limit: env_info.gas_limit.as_u64(),
+        coinbase: env_info.author,
+        number: U256::from(env_info.number),
+        timestamp: env_info.timestamp,
+        difficulty: env_info.difficulty,
+    }
+}
+
```
ouwenkg: `build_emv_info` may be better.

cita-executor/core/src/cita_executive.rs
```diff
@@ -256,50 +259,165 @@ impl<B: DB + 'static> CitaExecutive<B> {
...
+fn accrue_log(bloom: &mut Bloom, log: &LogEntry) {
+    // FIXME: upgrade ethereum-types to 0.6.0 to avoid unsafe block.
+    unsafe {
+        bloom.accrue(BloomInput::Raw(slice::from_raw_parts(
+            log.address.as_ptr(),
+            20,
+        )));
+    }
+
```
ouwenkg:
```rs
bloom.accrue(BloomInput::Raw(log.address.as_bytes()));
```
In m*, There must be a safe type converstion  from `address-> bytes`, maybe `copy_to` or somewhat.

cita-executor/core/src/cita_executive.rs
leeyr338: This is a function to build EVMContext, and `EVMContext` is defined in `cita-vm`. I think `Context` is okay for its meaning.

cita-executor/core/src/cita_executive.rs
leeyr338: Yes, have you seen my FIXME description?
```
// FIXME: upgrade ethereum-types to 0.6.0 to avoid unsafe block.
```

PR #695, opened 2019-07-18
[Feature] remove EVM: add call contract logic for CitaExecutive:exec

leeyr338:
- add create contract logic for CitaExecutive:exec.

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -12,6 +12,7 @@ use std::sync::Arc;
 use util::Bytes;
 
 use crate::authentication::{check_permission, AuthenticationError};
+use crate::contracts::native::factory::Factory as NativeFactory;
 use crate::core_types::{Bloom, BloomInput, Hash, LogEntry, Receipt, TypesError};
 use crate::libexecutor::economical_model::EconomicalModel;
 use crate::libexecutor::sys_config::BlockSysConfig;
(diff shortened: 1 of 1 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -33,23 +34,26 @@ const AMEND_KV_H256: u32 = 3;
...
     env_info: EnvInfo,
     economical_model: EconomicalModel,
 }
 
-impl<B: DB + 'static> CitaExecutive<B> {
+impl<'a, B: DB + 'static> CitaExecutive<'a, B> {
     pub fn new(
         block_provider: Arc<BlockDataProvider>,
         state_provider: State<B>,
+        native_factory: &'a NativeFactory,
```
ouwenkg: change to `native_provider`?

cita-executor/core/src/cita_executive.rs
```diff
@@ -154,11 +158,7 @@ impl<B: DB + 'static> CitaExecutive<B> {
                     data: t.data.clone(),
                 };
 
-                self.create(
-                    self.block_provider.clone(),
-                    self.state_provider.clone(),
-                    &params,
-                )
+                self.call_evm(&params)
```
ouwenkg: why  rename this? I think `create` may be better.

leeyr338: I want to reuse the code with `call` function.

cita-executor/core/src/cita_executive.rs
leeyr338: If you carefully taste, `Factory` and `provider` are different, I think it is more reasonable to use `Factory` here.

ouwenkg: OK，not a problem.

PR #699, opened 2019-07-20
Replace cita-database of cita-auth.

kaikai1024:
### Description of the Change

When i try to replace the db with cita-database which is implemented by rocksdb,
there is a package conflict of parity-rocksdb and rocksdb:

Screenshot from 2019-07-15 15-06-16

and sodiumoxide(something wrong when i try to upgrade it, check https://github.com/cryptape/cita-common/issues/281):

Screenshot from 2019-07-20 14-03-51

So i use a new parity-rocksdb branch that has the same unimplemented interface. then i can split the work into three parts:

* cita-aut
* cita-chain
* cita-executor

After all work are done, then we can use develop branch.

TODO: *The unit test.*

### DIFF
--- a/cita-auth/Cargo.toml
+++ b/cita-auth/Cargo.toml
@@ -21,13 +21,11 @@ libproto = { git = "https://github.com/cryptape/cita-common.git", branch = "deve
 cita-crypto = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
 tx_pool = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
 jsonrpc-types = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-core = { path = "../cita-chain/core" }
 uuid = { version = "0.7", features = ["v4"] }
 lru = "0.1"
 rayon = "1.0"
 hashable = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-db = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-evm = {path = "../cita-executor/evm"}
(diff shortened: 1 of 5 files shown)


cita-auth/src/main.rs
```diff
@@ -83,7 +82,7 @@ extern crate serde_derive;
 extern crate tempfile;
 #[macro_use]
 extern crate util;
-extern crate db as cita_db;
+extern crate cita_database;
```
leeyr338: You don't need to `extern` the crate in Edition 2018?

cita-auth/Cargo.toml
```diff
@@ -21,13 +21,11 @@ libproto = { git = "https://github.com/cryptape/cita-common.git", branch = "deve
 cita-crypto = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
 tx_pool = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
 jsonrpc-types = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-core = { path = "../cita-chain/core" }
 uuid = { version = "0.7", features = ["v4"] }
 lru = "0.1"
 rayon = "1.0"
 hashable = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-db = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-evm = {path = "../cita-executor/evm"}
```
kaikai1024: When i try to remove db, the evm imports the cita-common. so it's conflict with rockstdb.
Now it's not a problem.

cita-auth/src/main.rs
kaikai1024: DONE

PR #700, opened 2019-07-20
Remove snapshot. todo: rewrite it.

kaikai1024:
### Description of the Change

Will rewire the snapshot feature. 
So it's extra work for replacting cita-database. Remove it first.


leeyr338:
You just remove some files which related to DB. For the full feature of **Snapshot**, there are some other files to be deleted. But I think it okay for now, and please new an issue for it.

kaikai1024:
>  please new an issue for it.

Check teh #703

Delete it completely later if we need to.

PR #701, opened 2019-07-21
Use cita database chain

kaikai1024:
### Description of the Change

Rebase the branch: remove-snapshot. please review https://github.com/cryptape/cita/pull/700 first

### DIFF
--- a/cita-chain/Cargo.toml
+++ b/cita-chain/Cargo.toml
@@ -21,7 +21,7 @@ jsonrpc-types = { git = "https://github.com/cryptape/cita-common.git", branch = 
 core = { path = "./core" }
 common-types = { path = "./types" }
 proof = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-db = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
+cita_db = { git = "https://github.com/cryptape/cita-database.git", branch = "parity-rocksdb", package = "cita-database" }
 
 [build-dependencies]
 util = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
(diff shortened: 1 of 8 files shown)


cita-chain/src/main.rs
```diff
@@ -73,13 +73,14 @@ extern crate libproto;
 extern crate cita_logger as logger;
 #[macro_use]
 extern crate util;
-extern crate db as cita_db;
+
+extern crate cita_database as cita_db;
```
leeyr338: You don't need to use `extern` in Edition 2018?

cita-chain/src/main.rs
```diff
@@ -133,8 +134,8 @@ fn main() {
 
     let nosql_path = DataPath::nosql_path();
     trace!("nosql_path is {:?}", nosql_path);
-    let db_config = DatabaseConfig::with_columns(db::NUM_COLUMNS);
-    let db = Database::open(&db_config, &nosql_path).unwrap();
+    let db_config = DatabaseConfig::with_category_num(db::NUM_COLUMNS);
```
leeyr338: `COLUMN`  is a proper noun from **Storage**, so I think the name `with_columns` is better than `with_category_num`.

cita-chain/src/main.rs
```diff
@@ -133,8 +134,8 @@ fn main() {
...
+    let db = RocksDB::open(&nosql_path, &db_config).unwrap();
```
leeyr338: `expect` is better than `unwrap`, we need to know what happen when the system crash.

cita-chain/src/main.rs
kaikai1024: How the use `as` without `extern`?

cita-chain/src/main.rs
kaikai1024: You can check the cita-database. All the interface use the category as the params.
category is a abstract of column. this column is the noun fo rocksdb.

We can discuss it in a issue of cita-database.

cita-chain/src/main.rs
kaikai1024: It's anothere problem.

I just replace the old logic with cita-databse

cita-chain/src/main.rs
leeyr338: ok.

cita-chain/src/main.rs
leeyr338: ok

cita-chain/src/main.rs
leeyr338: Have you try:
```rust
use db as cita_db
```
?

cita-chain/src/main.rs
kaikai1024: #704

cita-chain/src/main.rs
kaikai1024: Wait a minute.

cita-chain/Cargo.toml
```diff
@@ -21,7 +21,7 @@ jsonrpc-types = { git = "https://github.com/cryptape/cita-common.git", branch =
 core = { path = "./core" }
 common-types = { path = "./types" }
 proof = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
-db = { git = "https://github.com/cryptape/cita-common.git", branch = "develop" }
+cita_db = { git = "https://github.com/cryptape/cita-database.git", branch = "parity-rocksdb", package = "cita-database" }
```
kaikai1024: @leeyr338  here

PR #707, opened 2019-07-23
Implement get function of cita-auth.

kaikai1024:
### Description of the Change

* Implement the get function of cita-auth
* Update the restore interface

### DIFF
--- a/cita-auth/src/txwal.rs
+++ b/cita-auth/src/txwal.rs
@@ -26,8 +26,7 @@ use std::sync::Arc;
 /// used to persist transaction pools message
 #[derive(Clone)]
 pub struct TxWal {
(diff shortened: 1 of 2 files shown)


cita-auth/src/txwal.rs
```diff
@@ -42,7 +41,7 @@ impl TxWal {
     pub fn regenerate(&mut self, path: &str) {
         let nosql_path = DataPath::root_node_path() + path;
         // FIXME
-        let _ = self.db.restore(&nosql_path, &String::from(path));
+        let _ = self.db.restore(&nosql_path);
```
ouwenkg: Remove this `FIXME`

kaikai1024: DONE

PR #724, opened 2019-07-26
[Feature] remove EVM: implement genesis.lazy_execute.[skip ci]

leeyr338:
- implement genesis.lazy_execute

### DIFF
--- a/cita-executor/core/src/libexecutor/executor.rs
+++ b/cita-executor/core/src/libexecutor/executor.rs
@@ -19,16 +19,15 @@ use super::command::{Command, CommandResp, Commander};
 use super::fsm::FSM;
 use super::sys_config::GlobalSysConfig;
 use crate::bloomchain::group::{BloomGroup, BloomGroupDatabase, GroupPosition};
-use crate::contracts::{native::factory::Factory as NativeFactory, solc::NodeManager};
+use crate::contracts::{solc::NodeManager};
 use crate::core::env_info::LastHashes;
-use crate::factory::*;
 use crate::header::*;
 pub use crate::libexecutor::block::*;
 use crate::libexecutor::genesis::Genesis;
 use crate::trie_db::TrieDB;
 use crate::types::ids::BlockId;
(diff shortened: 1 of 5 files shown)


cita-executor/core/src/libexecutor/executor.rs
```diff
@@ -66,28 +68,24 @@ impl Executor {
         // TODO: Can remove NUM_COLUMNS(useless)
         let config = Config::with_category_num(NUM_COLUMNS);
         let rocks_db = RocksDB::open(&data_path, &config).unwrap();
-        let trie_db = Arc::new(TrieDB::new(Arc::new(rocks_db)));
+        let db = Arc::new(rocks_db);
+        let trie_db = Arc::new(TrieDB::new(db.clone()));
```
ouwenkg: `state_db` might be better than `trie_db`, please check:

```rs
 pub fn lazy_execute(
        &mut self,	        &mut self,
        // state_db: &StateDB,	        state_db: Arc<CitaTrieDB>,
}
```

cita-executor/core/src/libexecutor/genesis.rs
```diff
@@ -99,83 +104,92 @@ impl Genesis {
...
+        .expect("Insert block header error.");
+        db.insert(
+            Some(DataCategory::Extra),
+            current_hash.to_vec(),
+            hash.clone(),
+        )
+        .expect("Insert block hash error.");
+        db.insert(Some(DataCategory::Extra), height, hash.clone())
+            .expect("Insert block hash error.");
+
```
ouwenkg: Do not use `expect()` here, It will panic when error occured.  Catch the db errors and return please.

leeyr338: I think can not insert a value to database is unacceptable， so I let it panic here.

cita-executor/core/src/libexecutor/executor.rs
leeyr338: Change the value-name trie_db to state_db.

PR #725, opened 2019-07-27
[Feature] remove EVM: fix unimplement in block_rs.[skip ci]

leeyr338:
- fix unimplement in block_rs.
- let Charge model unimplement.

### DIFF
--- a/cita-executor/core/src/libexecutor/block.rs
+++ b/cita-executor/core/src/libexecutor/block.rs
@@ -23,13 +23,12 @@ use crate::error::Error;
 // use crate::libexecutor::auto_exec::auto_exec;
 use crate::core::env_info::LastHashes;
 use crate::libexecutor::economical_model::EconomicalModel;
+use crate::libexecutor::executor::CitaTrieDB;
 use crate::libexecutor::sys_config::BlockSysConfig;
 use crate::receipt::{Receipt, ReceiptError};
-use crate::trie_db::TrieDB;
 use crate::tx_gas_schedule::TxGasSchedule;
 pub use crate::types::block::{Block, BlockBody, OpenBlock};
 use crate::types::transaction::SignedTransaction;
-use cita_database::RocksDB;
 use cita_merklehash;
 use cita_types::{Address, H256, U256};
 use cita_vm::BlockDataProvider;
@@ -37,6 +36,7 @@ use cita_vm::{evm::Error as EVMError, state::State as CitaState, Error as VMErro
 use hashable::Hashable;
 use libproto::executor::{ExecutedInfo, ReceiptWithOption};
 use rlp::*;
+use std::cell::RefCell;
 use std::cmp;
 use std::collections::HashMap;
 use std::ops::{Deref, DerefMut};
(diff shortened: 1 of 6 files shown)


cita-executor/core/src/libexecutor/block.rs
```diff
@@ -50,14 +50,14 @@ lazy_static! {
...
-    fn drain(self) -> TrieDB<RocksDB>;
+    /// Drop this object.
+    fn drain(self);
 }
 
 pub struct ExecutedBlock {
     pub block: OpenBlock,
     pub receipts: Vec<Receipt>,
-    pub trie_db: Arc<TrieDB<RocksDB>>,
+    pub state: Arc<RefCell<CitaState<CitaTrieDB>>>,
```
kaikai1024: Why use RefCell?

kaikai1024:
Better to open a new pr to use one unified word: `trie_db`, `state`, or `state_db`.

cita-executor/core/src/libexecutor/block.rs
```diff
@@ -50,14 +50,14 @@ lazy_static! {
 
 /// Trait for a object that has a state database.
 pub trait Drain {
-    /// Drop this object and return the underlieing database.
-    fn drain(self) -> TrieDB<RocksDB>;
+    /// Drop this object.
+    fn drain(self);
```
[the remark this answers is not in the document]
leeyr338: I checked the place where called it, this function just clear a cache, it do not need to return a DB object now.

cita-executor/core/src/libexecutor/block.rs
leeyr338: cita-vm expected a RefCell, and I should build the cell at the beginning of opening an `executedBlock`. 
We should keep only one stateDB for the whole block transactions.

leeyr338:
> Better to open a new pr to use one unified word: `trie_db`, `state`, or `state_db`.

Okay, I will do that later.

cita-executor/core/src/libexecutor/block.rs
kaikai1024: > We should keep only one stateDB for the whole block transactions.

one struct? there're three trie: state, tx, receipt,

cita-executor/core/src/libexecutor/block.rs
```diff
@@ -85,14 +85,20 @@ impl ExecutedBlock {
...
         conf: &BlockSysConfig,
         block: OpenBlock,
-        trie_db: Arc<TrieDB<RocksDB>>,
+        trie_db: Arc<CitaTrieDB>,
         state_root: H256,
         last_hashes: Arc<LastHashes>,
         eth_compatibility: bool,
     ) -> Result<Self, Error> {
+        let state = CitaState::from_existing(Arc::<CitaTrieDB>::clone(&trie_db), state_root)
+            .expect("Get state from trie db");
```
kaikai1024: Maybe `expect` is not a good idea. It'll panic. That's not what we want

leeyr338: But, if it cannot get state correctly, the logic can do nothing!

PR #734, opened 2019-07-29
fix insert operation

ouwenkg:
## Two things:

1. do rlp with the value
2. add missied update CurrentHeight process


leeyr338:
All `key`s for DB should be an index from `extras.rs`, please double check the code.

ouwenkg:
> All `key`s for DB should be an index from `extras.rs`, please double check the code.

Please check it again. thxs

PR #738, opened 2019-07-30
[Feature] remove EVM: Fix quota use for receipt.[skip travis]

leeyr338:
- Fix quota use for receipt

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -112,7 +112,7 @@ impl<'a, B: DB + 'static> CitaExecutive<'a, B> {
 
         let init_gas = t.gas - U256::from(base_gas_required);
 
-        match t.action {
+        let result = match t.action {
             Action::Store | Action::AbiStore => {
                 // Prepaid t.gas for the transaction.
                 self.prepaid(t.sender(), t.gas, t.gas_price, t.value)?;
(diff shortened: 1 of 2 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -525,16 +551,19 @@ fn build_result_with_ok(init_gas: U256, ret: InterpreterResult) -> ExecutedResul
 
     match ret {
         InterpreterResult::Normal(data, quota_left, logs) => {
-            result.quota_used = init_gas.as_u64() - quota_left;
+            result.quota_used = init_gas - U256::from(quota_left);
```
kaikai1024:
```
            result.quota_used = init_gas - U256::from(quota_left);
            result.quota_left = U256::from(quota_left);
```

These codes are nothing about `match ret`, better to put them outside?

cita-executor/core/src/cita_executive.rs
```diff
@@ -741,13 +770,18 @@ impl Default for EnvInfo {
 pub struct ExecutedResult {
     pub state_root: Hash,
     pub transaction_hash: Hash,
-    pub quota_used: u64,
+    pub quota_used: U256,
+    pub quota_left: U256,
```
kaikai1024: What `quota_left` be used for?

cita-executor/core/src/cita_executive.rs
```diff
@@ -741,13 +770,18 @@ impl Default for EnvInfo {
...
     pub logs_bloom: Bloom,
     pub logs: Vec<LogEntry>,
     pub exception: Option<ExecutionError>,
     pub contract_address: Option<Address>,
     pub account_nonce: U256,
 
+    // Note: if the transaction is a cita-evm call, needn't to handle the refund.
```
kaikai1024: should be `cita-vm`?

PR #744, opened 2019-07-30
Remove unused file

jerry-yu:
Remove unused file


cita-chain/types/src/lib.rs
```diff
@@ -43,7 +43,7 @@ pub mod log_blooms;
 pub mod log_entry;
 pub mod receipt;
 pub mod reserved_addresses;
-pub mod state_diff;
+//pub mod state_diff;
```
ouwenkg: Delete it please.

jerry-yu: maybe next time

PR #745, opened 2019-07-30
Fix: error key using in get_chain.

leeyr338:
Fix: error key using in get_chain.

### DIFF
--- a/cita-chain/core/src/libchain/chain.rs
+++ b/cita-chain/core/src/libchain/chain.rs
@@ -44,6 +44,7 @@ use cita_types::{Address, H256, H264, U256};
 use crate::cita_db::RocksDB;
 use crate::db::DBIndex;
 use cita_db::Database;
+use rlp::decode;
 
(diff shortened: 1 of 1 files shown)


cita-chain/core/src/libchain/chain.rs
```diff
@@ -306,14 +307,19 @@ pub fn get_chain(db: &RocksDB) -> Option<Header> {
             &CurrentHash.get_index().to_vec(),
         )
         .unwrap_or(None)
-        .map(|h| H256::from_slice(&h));
+        .map(|h| decode::<H256>(&h));
 
     if let Some(hash) = res {
+        trace!("Get block height from hash : {:?}", hash);
+        let hash_key = (&hash as &DBIndex<BlockNumber, Item = H256>).get_index();
```
kaikai1024: maybe we should define a unified key-transfer level to use the same rule of key in future.

leeyr338: okay

PR #755, opened 2019-08-01
Fix: Set state root error in close block, and add some traces.

leeyr338:
Fix: Set state root error in close block, and add some traces.

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -442,9 +442,14 @@ impl<'a, B: DB + 'static> CitaExecutive<'a, B> {
     }
 
     fn call_evm(&mut self, params: &VmExecParams) -> Result<ExecutedResult, ExecutionError> {
-        let evm_transaction = build_evm_transaction(params);
-        let evm_config = build_evm_config(self.env_info.gas_limit.as_u64());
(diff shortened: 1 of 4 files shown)


cita-executor/core/src/libexecutor/block.rs
```diff
@@ -334,7 +333,7 @@ impl ExecutedBlock {
         }
         // Rebuild block
         let mut block = Block::new(self.block);
-        let state_root = self.state_root;
+        let state_root = self.state.borrow_mut().root;
```
jerry-yu: use borrow()

cita-executor/core/src/cita_executive.rs
```diff
@@ -559,21 +564,31 @@ fn build_result_with_ok(init_gas: U256, ret: InterpreterResult) -> ExecutedResul
...
             result.output = data;
         }
-        InterpreterResult::Revert(_data, quota_left) => {
+        InterpreterResult::Revert(data, quota_left) => {
             result.quota_used = init_gas - U256::from(quota_left);
             result.quota_left = U256::from(quota_left);
+            trace!(
+                "Get data after executed the transaction [Revert]: {:?}",
+                data
+            );
```
ouwenkg: 打印这些 data 出来意义是什么呢，全是二进制，在日志里也很难看懂。

PR #757, opened 2019-08-01
fix create genesis tool using new vm

ouwenkg:


tools/create-genesis/Cargo.toml
```diff
@@ -9,14 +9,14 @@ serde = { version = "1.0", features = ["derive"] }
 serde_yaml = "0.8"
 json = "0.11.13"
 serde_json = "1.0.39"
-ethabi = "7.0.0"
-ethereum-types = "0.5.2"
+ethabi = "6.1.0"
+ethereum-types = "0.4"
```
leeyr338: why ethereum types should be `0.4` ? and ethabi should be `6.2.0`

tools/create-genesis/src/genesis.rs
```diff
@@ -1,17 +1,19 @@
+use std::collections::BTreeMap;
+use std::fs::File;
+use std::io::Write;
+use std::path::Path;
+use std::str::FromStr;
+
 use crate::common::{clean_0x, string_2_bytes};
```
leeyr338: we have a `clean_0x` function in cita_types, why not use it?

tools/create-genesis/src/genesis.rs
```diff
@@ -1,17 +1,19 @@
...
 use crate::contracts::ContractsData;
 use crate::miner::Miner;
 use crate::params::InitData;
 use crate::solc::Solc;
-use cita_types::U256;
+
 use ethabi::Contract;
+use ethereum_types::U256;
```
leeyr338: Can use the type `U256` in cita_types?

ouwenkg: Done

tools/create-genesis/src/genesis.rs
ouwenkg: Done

tools/create-genesis/Cargo.toml
ouwenkg: Done

tools/create-genesis/src/common.rs
```diff
@@ -1,51 +1,15 @@
-// use ethereum_types::Public;
-pub fn clean_0x(s: &str) -> &str {
-    if s.starts_with("0x") {
-        &s[2..]
-    } else {
-        s
-    }
-}
+// pub fn clean_0x(s: &str) -> &str {
```
kaikai1024: No need to keep these codes.

tools/create-genesis/Cargo.toml
ouwenkg: ethabi 7.0.0 -> ethereum-types 0.5.2
ethabi 6.1.0 -> ethereum-types 0.4

So i dump the ethabi version.

PR #761, opened 2019-08-02
Refine: error handle in cita_executive.

leeyr338:
- Refine: error handle in cita_executive.

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -1,7 +1,7 @@
 use cita_trie::DB;
 use cita_types::{Address, H160, H256, U256, U512};
 use cita_vm::{
-    evm::{Context as EVMContext, InterpreterResult, Log as EVMLog},
+    evm::{Context as EVMContext, Error as EVMError, InterpreterResult, Log as EVMLog},
     state::{Error as StateError, State, StateObjectInfo},
     BlockDataProvider, Config as VMConfig, DataProvider, Error as VMError, Store as VmSubState,
     Transaction as EVMTransaction,
@@ -15,7 +15,7 @@ use util::Bytes;
 
 use crate::authentication::{check_permission, AuthenticationError};
 use crate::contracts::native::factory::{Factory as NativeFactory, NativeError};
-use crate::core_types::{Bloom, BloomInput, Hash, TypesError};
+use crate::core_types::{Bloom, BloomInput, Hash};
 use crate::error::CallError;
 use crate::libexecutor::economical_model::EconomicalModel;
 use crate::libexecutor::sys_config::BlockSysConfig;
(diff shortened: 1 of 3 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -101,7 +101,7 @@ impl<'a, B: DB + 'static> CitaExecutive<'a, B> {
             //        got: U256,
             //    }
             // Need to change VMError defined in cita-vm.
-            return Err(ExecutionError::VM(VMError::NotEnoughBaseGas));
+            return Err(ExecutionError::NotEnoughBaseGas);
```
ouwenkg: remove the comment in last line.

cita-executor/core/src/cita_executive.rs
```diff
@@ -786,6 +793,41 @@ impl Default for EnvInfo {
...
+        ExecutedException::VM(err)
+    }
+}
+
+impl From<NativeError> for ExecutedException {
+    fn from(err: NativeError) -> Self {
+        ExecutedException::NativeContract(err)
+    }
+}
+
```
ouwenkg: Put all error handling logic in another seperated file is better.

cita-executor/core/src/cita_executive.rs
```diff
@@ -639,56 +651,51 @@ pub fn contract_address(address: &Address, nonce: &U256) -> Address {
...
+    /// Return when authentication error occurs.
     Authentication(AuthenticationError),
-    NativeContract(NativeError),
+    /// Return when the quota_limit in transaction lower then base quota required.
+    NotEnoughBaseGas,
+    /// Return when transaction nonce does not match state nonce.
+    InvalidNonce,
+    /// Return when the cost of transaction (value + quota_price * quota) exceeds.
+    NotEnoughBalance,
+    /// Return when the block quota exceeds block quota limit.
```
ouwenkg: I am a little confused about what the difference between `ExecutionError` and `ExecutedException`.

`ExecutedException` wraps error returned from vm?  if this logic is right,  Maybe `InvalidNonce` and `NotEnoughBalance` belongs to `ExecutedException`.

leeyr338: `ExecutedException` will be set into `Receipt`, but `ExecutionError` will be handled in `block.rs` where called the `exec` function.

cita-executor/core/src/cita_executive.rs
leeyr338: This is some errors about `executive`, I will think about it later!

cita-executor/core/src/cita_executive.rs
leeyr338: Why? This is a FIXME comment, and I also think the VMError should be redefined.

PR #764, opened 2019-08-05
fix amend test

ouwenkg:
```sh
cd cita-executor/

cargo test -p core-executor

```

选区_035


ouwenkg:
Retain `get` functionality temporatry , then refacor amend process later and add option label `#[test]`

PR #766, opened 2019-08-05
Fix executor unit test.

kaikai1024:
### Description of the Change

`cd cita-executor && cargo test -p cita-executor`

Screenshot from 2019-08-05 11-22-47

### DIFF
--- a/cita-executor/src/backlogs.rs
+++ b/cita-executor/src/backlogs.rs
@@ -414,12 +414,11 @@ pub fn wrap_height(height: usize) -> u64 {
 #[cfg(test)]
 mod tests {
     use super::{wrap_height, Backlog, Backlogs, Priority};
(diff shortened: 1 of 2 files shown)


cita-executor/src/backlogs.rs
```diff
@@ -494,19 +494,21 @@ mod tests {
     #[test]
     fn test_backlog_is_completed_with_none() {
         let height = 9;
+        let path = "test-rocksdb/backlog_is_completed_with_none";
```
ouwenkg: It would be better to define a const variable point to `test-rocksdb`

kaikai1024: yeah. 
btw, It'll be useless in the future

```
// TODO Use MemoryDB instead of RocksDB.
```

PR #771, opened 2019-08-05
rewrite ids under cita chain types

ouwenkg:


cita-chain/types/src/block_tag.rs
```diff
@@ -0,0 +1,25 @@
...
+    Hash(H256),
+    Number(BlockNumber),
+    Earliest,
+    Latest,
+    Pending,
+}
+
+impl From<RpcBlockNumber> for BlockTag {
+    fn from(v: RpcBlockNumber) -> BlockTag {
+        match v {
```
leeyr338: what means for `v` ?

ouwenkg: OK, it should be `n` or `rpc_block_number`.

kaikai1024:
Rename `BlockId` to `BlockTag`?

*I'm not sure it's a good idea. Just worried about if there is license problem in future.*

ouwenkg:
> Rename `BlockId` to `BlockTag` to avoid GPL?
> 
> _I'm not sure it's a good idea. Just worried about if there is license problem in future._

What you worry about exits indeed, any advices about this now?

kaikai1024:
Better than none.

*btw, Use `rename` is better than `rewrite`.*

cita-chain/types/src/block_number.rs
```diff
@@ -0,0 +1,30 @@
+use cita_types::H256;
+use jsonrpc_types::rpc_types::{BlockNumber as RpcBlockNumber, BlockTag as RpcBlockTag};
+
+pub type TransactionHash = H256;
+pub type BlockNumber = u64;
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum BlockTag {
```
leeyr338: I feel so strange about this define!
Do you think `Height` and `Hash` is a `Tag`?
If `yes`, it should be included in `Tag` enum.
If `not`, the enum name should not be `BlockTag`.

leeyr338: I don't have a better idea for now.

ouwenkg: At present, just let it go. More modified later.

PR #772, opened 2019-08-05
remove db.rs in chain

ouwenkg:


leeyr338:
extras.rs is also a GPL file, you need to double check about this file.

ouwenkg:
> extras.rs is also a GPL file, you need to double check about this file.

Yeal, next pr today.

PR #774, opened 2019-08-06
Fix: main process exit in chain.

leeyr338:
- Fix: main process exit in chain.

### DIFF
--- a/cita-chain/src/main.rs
+++ b/cita-chain/src/main.rs
@@ -154,32 +154,30 @@ fn main() {
     });
 
     // Write: add block
-    thread::spawn(move || {
-        let mut timeout_factor = 0u8;
-        loop {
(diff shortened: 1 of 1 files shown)


leeyr338:
> How?

Keep an infinite loop for main.

kaikai1024:
> Keep an infinite loop for main.

Then?  That's the old `cita-chain` logic, what changed?

ouwenkg:
> > Keep an infinite loop for main.
> 
> Then? That's the old `cita-chain` logic, what changed?

Chain process panic caused by previous remove cache issue.  I mistaked del loop with cache collection.

kaikai1024:
> Chain process panic caused by previous remove cache issue. I mistaked del loop with cache collection.

Got it.

PR #781, opened 2019-08-06
split extras

ouwenkg:
- Taking `BlockReceipts` and `TransactionIndex` out firstly.


cita-chain/types/src/block_receipts.rs
```diff
@@ -0,0 +1,45 @@
+use crate::receipt::Receipt;
```
kaikai1024: Add a tag `FIXME` to rewrite later?

PR #787, opened 2019-08-08
unique error.rs

ouwenkg:
### Changes

- Del `error.rs` in executor and retain only `error.rs` in chain.
- Relocated `error.rs` in `chain-core` to `chain-typtes`.


cita-executor/core/src/libexecutor/command.rs
```diff
@@ -24,7 +24,7 @@ use crate::contracts::native::factory::Factory as NativeFactory;
 use crate::contracts::solc::{
     sys_config::ChainId, PermissionManagement, SysConfig, VersionManager,
 };
-use crate::error::CallError;
+use crate::executed::CallError;
```
kaikai1024: Is it the same error?

ouwenkg:
```
// error.rs
pub use crate::executed::{CallError, ExecutionError};
```

PR #789, opened 2019-08-08
fix fee and error

jerry-yu:
fix gas for error

### DIFF
--- a/cita-executor/core/src/cita_executive.rs
+++ b/cita-executor/core/src/cita_executive.rs
@@ -1,22 +1,23 @@
 use cita_trie::DB;
 use cita_types::{Address, H160, H256, U256, U512};
+use cita_vm::Error as ExecErr;
 use cita_vm::{
     evm::{Context as EVMContext, Error as EVMError, InterpreterResult, Log as EVMLog},
-    state::{Error as StateError, State, StateObjectInfo},
-    BlockDataProvider, Config as VMConfig, DataProvider, Error as VMError, Store as VmSubState,
+    state::{State, StateObjectInfo},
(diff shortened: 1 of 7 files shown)


cita-executor/core/src/cita_executive.rs
```diff
@@ -230,37 +232,44 @@ impl<'a, B: DB + 'static> CitaExecutive<'a, B> {
                         ExecutedException::VM(VMError::NotEnoughBalance) => {
                             return Err(ExecutionError::NotEnoughBalance)
                         }
+                        ExecutedException::VM(VMError::NotEnoughBalance) => {
+                            return Err(ExecutionError::NotEnoughBalance)
+                        }
```
ouwenkg: 这一段为什么注释掉了。

jerry-yu: 这段ok的，不用处理

cita-executor/core/src/libexecutor/block.rs
```diff
@@ -217,43 +189,66 @@ impl ExecutedBlock {
...
-                let tx_quota_used = match err {
-                    ExecutionError::Internal(_) => t.gas,
-                    _ => cmp::min(
-                        self.state
-                            .borrow_mut()
-                            .balance(&sender)
-                            .unwrap_or_else(|_| U256::from(0)),
-                        U256::from(schedule.tx_gas),
-                    ),
+                let tx_quota_used = match receipt_error {
```
leeyr338: Delete the `cmp::min` logic will lead incompatible, the logic is different as before.

PR #796, opened 2019-08-12
remove license docs about log_blooms.rs

ouwenkg:


cita-executor/core/src/libexecutor/executor.rs
```diff
@@ -388,22 +386,6 @@ impl Executor {
...
-            .get(Some(DataCategory::Extra), &*position.get_index())
-            .map(|bytes| decode::<LogBloomGroup>(bytes.unwrap().as_slice()))
-            .expect("Get bloom group from db error.");
-
-        let bloom_group = log_bloom_group.into();
-
-        Some(bloom_group)
-    }
-}
-
```
ouwenkg: Another implementation already existed in [cita-chain](https://github.com/cryptape/cita/blob/develop/cita-chain/core/src/libchain/chain.rs#L257). For reviewers.

jerry-yu: add ours licence ？

cita-chain/types/src/header.rs
```diff
@@ -34,7 +34,7 @@ use hashable::{Hashable, HASH_NULL_RLP};
 use util::HeapSizeOf;
 
 // from parity
-pub use crate::log_blooms::LogBloomGroup;
+// pub use crate::log_blooms::LogBloomGroup;
```
leeyr338: Just delete this line?

cita-executor/core/src/libexecutor/executor.rs
ouwenkg: I think there is no need to add such supplement docs.

In my respective, all the part of license docs in files should be deleted after we finish removing gpl. No matter it is `parity` or `cryptape`

cita-chain/types/src/header.rs
ouwenkg: I update again.

PR #800, opened 2019-08-13
remove executed.rs

ouwenkg:


cita-executor/core/README.md
```diff
@@ -1,44 +1 @@
 # Executor microservice core for CITA
```
kaikai1024: Remove the file directly?

ouwenkg: Done

PR #801, opened 2019-08-13
Chores

ouwenkg:
### Description of the Change

* Remove useless crate in chain and executor
* Remove `core-types` in executor
* Deal with some `unwrap()`
* Remove useless `extern`


cita-chain/core/Cargo.toml
```diff
@@ -1,22 +1,21 @@
 [package]
 name = "core"
 version = "0.1.0"
-authors = ["Cryptape Technologies <contact@cryptape.com>", "Parity Technologies <admin@parity.io>"]
+authors = ["Cryptape Technologies <contact@cryptape.com>"]
```
kaikai1024: There'are still codes about `filter` of `Parity`.

But, it's fine.

ouwenkg: Just let it go.

Issue #675, opened 2019-07-01
chainId changed unexpected while upgrading 0.19rc to 0.20


ouwenkg:
## Description

模拟测试chainId 为3，0.19rc 链，升级0.20。检测是否出现 chainId 异常，操作脚本如下：
 
https://gist.github.com/ouwenkg/dc7825ed58cd02d3bad4548f59226b3f

Output:
```
1. Prepare 0.19rc envionoment(compiled from 0.19rc tag)
2. Change default chainId(modify chainId setting in scripts/)
3. Run 0.19rc chain firstly
Initializing chain from provided state
INFO:cita.block	Block pre-sealed, 2162077 quota used 
INFO:cita.chain	Adding to head head=4c29c1f1
Saved 3 address change logs
INFO:cita.chain	Added block 1 (4b0d01cf) with 1 txs and 2162077 quota 
INFO:cita.block	Block pre-sealed, 2357091 quota used 
INFO:cita.chain	Adding to head head=4b0d01cf
Saved 2 address change logs
INFO:cita.chain	Added block 2 (dd235979) with 1 txs and 2357091 quota 
INFO:cita.block	Block pre-sealed, 1035750 quota used 
INFO:cita.chain	Adding to head head=dd235979
Saved 2 address change logs
INFO:cita.chain	Added block 3 (f6c5e1b7) with 1 txs and 1035750 quota 
INFO:cita.block	Block pre-sealed, 2073590 quota used 
INFO:cita.chain	Adding to head head=f6c5e1b7
Saved 2 address change logs
INFO:cita.chain	Added block 4 (1011324d) with 1 txs and 2073590 quota 
INFO:cita.block	Block pre-sealed, 2657330 quota used 
INFO:cita.chain	Adding to head head=1011324d
Saved 2 address change logs
INFO:cita.chain	Added block 5 (8552ed0f) with 1 txs and 2657330 quota 
INFO:cita.block	Block pre-sealed, 2942019 quota used 
INFO:cita.chain	Adding to head head=8552ed0f
Saved 2 address change logs
INFO:cita.chain	Added block 6 (922b6742) with 1 txs and 2942019 quota 
INFO:cita.block	Block pre-sealed, 4292430 quota used 
INFO:cita.chain	Adding to head head=922b6742
Saved 2 address change logs
INFO:cita.chain	Added block 7 (e55700d6) with 1 txs and 4292430 quota 
INFO:cita.block	Block pre-sealed, 2292209 quota used 
[...]

PR #806, opened 2019-08-15
Fix the script of amend system contracts.

kaikai1024:
### Description of the Change

The script should use the param to generate the genesis file.

### Applicable Issues

#675


kaikai1024:
Reopen it after the CI be fixed.

PR #812, opened 2019-08-23
Rewrite log & blooms.

kaikai1024:
### Description of the Change

* Refactor the filter type used by chain
* Rename log_entry to log
* Chore about log
* Add unit tests for log


cita-chain/types/src/header.rs
```diff
@@ -33,8 +33,6 @@ pub use crate::block_number::BlockNumber;
 use hashable::{Hashable, HASH_NULL_RLP};
 use util::HeapSizeOf;
 
-// from parity
-pub use crate::log_entry::LogBloom;
 lazy_static! {
     pub static ref ZERO_LOGBLOOM: LogBloom = LogBloom::from([0x00; 256]);
 }
```
ouwenkg: `Line 39` has GPL suspicion.

cita-chain/types/src/log.rs
```diff
@@ -0,0 +1,193 @@
...
+        })
+    }
+}
+
+impl HeapSizeOf for Log {
+    fn heap_size_of_children(&self) -> usize {
+        self.topics.heap_size_of_children() + self.data.heap_size_of_children()
+    }
+}
+
```
ouwenkg: `HeapSizeOf` is useless now.

cita-chain/types/src/header.rs
kaikai1024: ok

cita-chain/types/src/log.rs
kaikai1024: ok, I'll have a try

PR #814, opened 2019-08-23
rpc test chores

ouwenkg:
### Description of the Change

* Update `cita-vm` version to latest in Cargo.lock
* Handle returned evm error `StackUnderflow`
* Set `block_quota_limit` in `apply_transaction()`
* Some test script refactor
* Add necessary log record in critical point

### DIFF
--- a/cita-executor/core/src/libexecutor/block.rs
+++ b/cita-executor/core/src/libexecutor/block.rs
@@ -39,6 +39,7 @@ use crate::libexecutor::auto_exec::auto_exec;
 use crate::libexecutor::economical_model::EconomicalModel;
 use crate::libexecutor::executor::CitaTrieDB;
 use crate::libexecutor::sys_config::BlockSysConfig;
+use crate::libexecutor::sys_config::GlobalSysConfig;
 use crate::receipt::Receipt;
 use crate::tx_gas_schedule::TxGasSchedule;
 pub use crate::types::block::{Block, BlockBody, OpenBlock};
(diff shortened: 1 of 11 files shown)


cita-executor/core/src/libexecutor/block.rs
```diff
@@ -137,9 +138,12 @@ impl ExecutedBlock {
         }
     }
 
-    #[allow(unknown_lints, clippy::too_many_arguments)] // TODO clippy
-    pub fn apply_transaction(&mut self, t: &SignedTransaction, conf: &BlockSysConfig) {
+    pub fn apply_transaction(&mut self, t: &SignedTransaction, sys_config: &GlobalSysConfig) {
```
kaikai1024: Why change the `BlockSysConfig` to `GlobalSysConfig`

cita-executor/core/src/libexecutor/command.rs
```diff
@@ -283,6 +282,7 @@ impl Commander for Executor {
             block_quota_limit: *header.quota_limit(),
             account_quota_limit: u64::max_value().into(),
         };
+        context.block_quota_limit = U256::from(self.sys_config.block_quota_limit);
```
kaikai1024: It's so weird. Why not just set the block_quota_limit at the first.

cita-executor/core/src/libexecutor/executor.rs
```diff
@@ -351,11 +350,11 @@ impl Executor {
     pub fn executed_result_by_height(&self, height: u64) -> ExecutedResult {
         let block_tag = BlockTag::Height(height);
         let sys_config = GlobalSysConfig::load(&self, block_tag);
-        let consensus_config = make_consensus_config(sys_config);
+        let consensus_config = make_consensus_config(sys_config.clone());
```
kaikai1024: Why add a clone?

cita-executor/core/src/libexecutor/command.rs
ouwenkg: The `block_quota_limit` in context from header is a default value. It should be updated. at first?

cita-executor/core/src/libexecutor/executor.rs
ouwenkg: useless, I remove it.

cita-executor/core/src/libexecutor/block.rs
ouwenkg: I need to update `block_quota_limit` next.

PR #815, opened 2019-08-23
use new bft

jerry-yu:
use new bft


kaikai1024:
No need to change the `.gitmodules`?

kaikai1024:
I'm afraid there will be a build problem. because the rocksdb and parity-rocksdb.
You can rebase the `remove-gpl` branch or merge the v0.3 into the remove-gpl.

PR #820, opened 2019-08-26
Chores about header.rs

ouwenkg:


cita-chain/types/src/header.rs
```diff
@@ -31,29 +31,28 @@ use time::get_time;
...
     /// Transactions root.
     transactions_root: H256,
-    /// Block gas limit.
+    /// Block quota limit.
     quota_limit: U256,
-    /// the proof of the block
+    /// Block proof
     proof: ProtoProof,
-    /// The version of the header.
+    /// Block version.
```
kaikai1024: Block version?

ouwenkg: Is there something wrong?

kaikai1024: We have `protocol version` and `software version`.
I don't think that `block version` is a good description.

PR #821, opened 2019-08-26
Remove useless HeapSizeOf

ouwenkg:


ouwenkg:
> when considering add cache, this maybe useful

That might be useful while designing our own cache implemention in the future.

This trait comes from outsite GPL crate. For now,  It is a must-done task to remove those code.

PR #824, opened 2019-08-26
fix sync block error

jerry-yu:
fix sync block error when in specific height 21

### DIFF
--- a/cita-executor/src/postman.rs
+++ b/cita-executor/src/postman.rs
@@ -305,7 +305,7 @@ impl Postman {
                 for proto_block in sync_res.take_blocks().into_iter() {
                     let open_block = OpenBlock::from(proto_block);
                     if !self.backlogs.insert_synchronized(open_block) {
-                        return false;
+                        continue;
                     }
                 }
                 true


cita-executor/src/postman.rs
```diff
@@ -305,7 +305,7 @@ impl Postman {
                 for proto_block in sync_res.take_blocks().into_iter() {
                     let open_block = OpenBlock::from(proto_block);
                     if !self.backlogs.insert_synchronized(open_block) {
-                        return false;
+                        continue;
```
keroro520: Could you add some more description for this PR?
What the problem is if `return false` here.

And can we add a test to reproduce this problem?

jerry-yu: you can try when node0's height is 21, node1 sync block from node0.
when node1's height reach 20, network always feed executor [20,21] block,but inserting block 20 return false,so block 21 can't be executed

keroro520: > when node1's height reach 20, network always feed executor [20,21] block

Why `network` doesn't feed `executor` `[21]` when retry?

I am curious what's preventing adding a test to reproduce this problem?

keroro520: Another question: why `21` is special?

jerry-yu: that is another question. here,why not deal with next block, just because of prev block duplicated

keroro520:
I will dismiss my RequestChange.
But I cannot ensure that the so-called 21-problem is resulted from "return early for duplicated". So I will remove myself from reviewers.

jerry-yu:
from whatever view, directly returning false is risky  and not friendly for optimize

PR #827, opened 2019-08-27
Change GPL to apache 2.0

leeyr338:
Change GPL to apache 2.0.

* Change LICENSE from GPL to Apache 2.0.
* Add license announce to Cargo.toml.
* Add license announce to each rust file.


cita-chain/types/src/receipt.rs
```diff
@@ -1,18 +1,16 @@
-// Copyright 2015-2017 Parity Technologies (UK) Ltd.
```
kaikai1024: Merge this PR after this file is replaced?

leeyr338: You mean we can not change this file to apache license ?

leeyr338: You mean we can not change this file to apache license for now ?

kaikai1024: Not Professional.
The project muse be GPL even there is only one GPL file accoring the GPL license.
Just an advice, yes, we can. but i can not apporve.

leeyr338: > The project muse be GPL even there is only one GPL file accoring to the GPL license.

Yes!

PR #830, opened 2019-08-28
[skip audit]fix state test

jerry-yu:
fix state test

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,4 +25,4 @@
 [submodule "tests/jsondata"]
 	path = tests/jsondata
 	url = https://github.com/cryptape/cita-testdata.git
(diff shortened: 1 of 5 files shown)


.gitmodules
```diff
@@ -25,4 +25,4 @@
 [submodule "tests/jsondata"]
 	path = tests/jsondata
 	url = https://github.com/cryptape/cita-testdata.git
-	branch = master
+	branch = master
```
ouwenkg: useless space.

ouwenkg:
I think we should raise this pr to `remove-gpl` at present. Updating submodule at develop branch may cause trouble some conflict while merging `remove-gpl` to `develop` in the future.

PR #841, opened 2019-08-30
try to add license in readme

jerry-yu:
try to add license in readme


jerry-yu:
wait for fossa completed

PR #842, opened 2019-09-02
use develop branch and fix clippy

jerry-yu:
use develop branch and fix clippy

### DIFF
--- a/cita-network/src/node_manager.rs
+++ b/cita-network/src/node_manager.rs
@@ -286,6 +286,8 @@ impl NodesManager {
         }
     }
 
+    // clippy
+    #[allow(clippy::drop_copy, clippy::zero_ptr)]
     pub fn run(&mut self) {
         loop {
             select! {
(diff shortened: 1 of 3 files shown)


cita-network/src/node_manager.rs
```diff
@@ -286,6 +286,8 @@ impl NodesManager {
         }
     }
 
+    // clippy
+    #[allow(clippy::drop_copy, clippy::zero_ptr)]
```
kaikai1024: Why not fix it?

jerry-yu: for clippy

jerry-yu: maybe you can try it first

jerry-yu: it is up to lib's macro select!

jerry-yu: it is due to use new crossbeam-channel and the reson for use crossbeam-channel,is upgrade cita-logger in bft

kaikai1024: ok

PR #845, opened 2019-09-03
Remove ec

kaikai1024:
### Description of the Change
* Add travis.
* Upgrade cita-sdk-js to avoid the ec dependence.


kaikai1024:
If https://github.com/cryptape/cita/pull/839 is merged first, then i will change the rarget branch to develop.

PR #847, opened 2019-09-03
Compile sm2

jerry-yu:
fix for compile sm2 not right

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -779,7 +779,7 @@ dependencies = [
 [[package]]
 name = "cita-vm"
 version = "0.2.1"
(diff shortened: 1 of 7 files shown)


Cargo.lock
```diff
@@ -2389,7 +2389,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "num-integer 0.1.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-traits 0.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "rand 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.22 (registry+https://github.com/rust-lang/crates.io-index)",
```
leeyr338: Must we use this lower version for `rand`?

PR #855, opened 2019-09-11
Interface block interval

ouwenkg:
### Description of the Change

After this pr,  the block interval can be changed according to your demand. 

Default block interval is 3s as before.

### DIFF
--- a/tests/integrate_test/cita_features_test.sh
+++ b/tests/integrate_test/cita_features_test.sh
@@ -65,7 +65,9 @@ main() {
 
     echo -e "6) Run block interval tests ...\n"
     test_change_block_interval 2
-    timeout=$(check_height_growth_normal 0 60) || (echo "Block interval changed")
+    timeout=$(check_height_growth_normal 0 60) && (echo "FAILED"
+                                                    echo "failed to change block interval"
+                                                    exit 1)
     echo "Done"
 }
 
--- a/tests/integrate_test/cita_features_test.sh
+++ b/tests/integrate_test/cita_features_test.sh
@@ -23,6 +23,13 @@ test_perm_denied() {
     cd ../../..
 }
 
+test_change_block_interval() {
+    local version=$1
+    cd ./scripts/txtool/txtool
+    python3 "${SOURCE_DIR}"/tests/integrate_test/test_block_interval.py --version="$version"
+    cd ../../..
+}
+
 main() {
     echo -n "0) prepare  ...  "
     # shellcheck source=/dev/null
@@ -54,6 +61,11 @@ main() {
 
     echo -n "5) Check permission denied ... "
     test_perm_denied 2
+    echo "Done"
+
+    echo -e "6) Run block interval tests ...\n"
+    test_change_block_interval 2
+    timeout=$(check_height_growth_normal 0 60) || (echo "Block interval changed")
     echo "Done"
 }
 
--- a/tests/integrate_test/test_block_interval.py
+++ b/tests/integrate_test/test_block_interval.py
@@ -21,7 +21,6 @@ def main():
 
     tx_hash = send_tx(
         admin_privkey, to=system_contract_address, code=code, version=version)
-    time.sleep(5)       # ensure tx execution.
     receipt = get_receipt(tx_hash)
     assert receipt['errorMessage'] == None, "Transaction execution failed"
 
--- a/tests/integrate_test/test_block_interval.py
+++ b/tests/integrate_test/test_block_interval.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+import argparse
+import time
+from txtool_utils import send_tx, get_receipt, get_metadata
+
+
+def main():
+    print(f'a) Test default block interval...')
+    metadata = get_metadata()
+    assert metadata['blockInterval'] == 3000
+
+    print(f'b) Send transaction to change block interval...')
+    # send tx to change the block interval
+    admin_privkey = '0x5f0258a4778057a8a7d97809bd209055b2fbafa654ce7d31ec7191066b9225e6'
+    system_contract_address = '0xffffffffffffffffffffffffffffffffff020000'
+    code = 'de7aa05d000000000000000000000000000000000000000000000000000000003b9aca00'  # setBlockInterval(1000000000)
+
+    opts = parse_arguments()
+    version = opts.version
+
+    tx_hash = send_tx(
+        admin_privkey, to=system_contract_address, code=code, version=version)
+    time.sleep(5)       # ensure tx execution.
+    receipt = get_receipt(tx_hash)
+    assert receipt['errorMessage'] == None, "Transaction execution failed"
+
+    metadata = get_metadata()
+    assert metadata['blockInterval'] == 1000000000, "Block interval not changed"
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--version", help="Tansaction version.", type=int)
+
+    return parser.parse_args()
+
+
+if __name__ == '__main__':
+    main()
--- a/scripts/contracts/src/system/SysConfig.sol
+++ b/scripts/contracts/src/system/SysConfig.sol
@@ -109,6 +109,13 @@ contract SysConfig is ISysConfig, EconomicalType, ReservedAddrPublic {
         chainName = _chainName;
     }
 
+    function setBlockInterval(uint64 _blockInterval)
+        external
+        onlyAdmin
+    {
+        blockInterval = _blockInterval;
+    }
+
     function updateToChainIdV1()
         external
         onlyAdmin
--- a/tests/integrate_test/txtool_utils.py
+++ b/tests/integrate_test/txtool_utils.py
@@ -60,3 +60,8 @@ def rpc_request(method, params):
 def get_balance(addr):
     """ Get the balance of an address """
     return int(rpc_request('getBalance', [addr, 'pending']), 16)
+
+def get_metadata():
+    """ Get metadata """
+    response = rpc_request('getMetaData', ["pending"])
+    return response


tests/integrate_test/test_block_interval.py
```diff
@@ -0,0 +1,59 @@
...
+    time.sleep(6)
+    receipt = get_receipt(tx_hash)
+    assert receipt['errorMessage'] == None, "Transaction execution failed"
+
+    print(f'c) Test changed block interval...')
+    current_height = get_current_block_number()
+    height = int(current_height, 16)
+    next_height = str(hex(height + 1))
+
+    time.sleep(3)
```
kaikai1024: IMHO, This is a very unstable test.

tests/integrate_test/test_block_interval.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # block interval is 6s now, you could not get next block 3s after.
+    next_block = get_block_by_number(next_height)
+    assert next_block == None, "Block interval had changed to 6s"
```
leeyr338: I don't quite understand why you are saying "Block interval had changed to 6s" here ?

ouwenkg:
> `cita-bft` will receive the changed `block_interval` value?

Yes， chain will send `RichStatus` which block_internal included to bft when chain grows.

tests/integrate_test/test_block_interval.py
ouwenkg: A little..., may be just check  metadata is ok.

kaikai1024: You can try to set the `block interval` value to a big number. then use the `check_height_stopped` with a resonable timeout.

tests/integrate_test/test_block_interval.py
ouwenkg: I send a tx to invoke `setBlockInterval`, and changed the default block interval from 3s to 6s.

kaikai1024:
> chain will send RichStatus which block_internal included to bft when chain grows.

ok.

```rust
pub struct RichStatus {
    number: u64,
    hash: H256,
    nodes: Vec<Address>,
}
```
btw, `number` is a very bad name..

tests/integrate_test/test_block_interval.py
leeyr338: Yes, you have sent the transaction to `setBlockInterval`, but you can not say:
```
The blockchain has works as exactly 6s for its interval. 
```
from your test case, right?

As an option, you can count for more times, and every time, the interval is about 6s. And this can high probability say "It have really to change to 6s interval" .

ouwenkg:
> > chain will send RichStatus which block_internal included to bft when chain grows.
> 
> ok.
> 
> ```rust
> pub struct RichStatus {
>     number: u64,
>     hash: H256,
>     nodes: Vec<Address>,
> }
> ```
> 
> btw, `number` is a very bad name..

```proto
message RichStatus {
    bytes hash = 1;
    uint64 height = 2;
    repeated bytes nodes = 3;
    uint64 interval = 4;
    uint32 version = 5;
    repeated bytes validators = 6;
}
```
This is the whole mq-msg  `RichStatus` structure.

kaikai1024:
> This is the whole mq-msg RichStatus structure.

That's what i mean. why not use `block interval` or `interval`, How can i know the `number` is what number when i read the struct `RIchStatus`.

That's another problem.

Forget it.

tests/integrate_test/test_block_interval.py
ouwenkg: A stable test case about this is hard to design. 

In my view, the test is successfully as long as the default block interval changed after tx, i do not care how long the block interval you had set in tx.
 
In other words, I could not get a block after 3 seconds from now that shows block interval had changed. As for the how block interval changed, check it by `getMetaDta` is ok too.

tests/integrate_test/test_block_interval.py
ouwenkg: Please check again. I do some modifications to make this test stable.

tests/integrate_test/test_block_interval.py
```diff
@@ -0,0 +1,62 @@
...
+    time.sleep(5)       # ensure tx execution.
+    receipt = get_receipt(tx_hash)
+    assert receipt['errorMessage'] == None, "Transaction execution failed"
+
+    print(f'c) Test changed block interval...')
+    current_height = get_current_block_number()
+    height = int(current_height, 16)
+    next_height = str(hex(height + 1))
+
+    time.sleep(5)
```
leeyr338: 这个用例使用了几个 sleep ，其结果判断是有随机风险的。举个例子：
如果在 37 行获得高度的时刻，刚好上一次出块后进行了 6s ，而你这里 sleep(5) 之后，再去 get_block_by_number，实际上是可以获得 next_block 的。
在这种情况下，这个用例是失败的。

tests/integrate_test/test_block_interval.py
kaikai1024: But `sleep` is the real reason of unstable not just `3s`.

tests/integrate_test/test_block_interval.py
ouwenkg: 我换了个思路，不取块内容了，用固定时间段内的高度差来做为判断依据。

kaikai1024:
And you need to consider the upgrade.

tests/integrate_test/cita_features_test.sh
```diff
@@ -55,6 +62,12 @@ main() {
     echo -n "5) Check permission denied ... "
     test_perm_denied 2
     echo "Done"
+
+    echo -e "6) Run block interval tests ...\n"
+    test_change_block_interval 2
+    timeout=$(check_height_growth_normal 0 60) || (echo "Block interval changed"
```
leeyr338: 你这样写，用例肯定测试不通过啊。

ouwenkg:
> And you need to consider the upgrade.

Another pr would be better.

kaikai1024:
> Another pr would be better.

yes, just FYI.

tests/integrate_test/test_block_interval.py
```diff
@@ -0,0 +1,40 @@
...
+    admin_privkey = '0x5f0258a4778057a8a7d97809bd209055b2fbafa654ce7d31ec7191066b9225e6'
+    system_contract_address = '0xffffffffffffffffffffffffffffffffff020000'
+    code = 'de7aa05d000000000000000000000000000000000000000000000000000000003b9aca00'  # setBlockInterval(1000000000)
+
+    opts = parse_arguments()
+    version = opts.version
+
+    tx_hash = send_tx(
+        admin_privkey, to=system_contract_address, code=code, version=version)
+    time.sleep(5)       # ensure tx execution.
```
jerry-yu: Checking Block height increase by 2 , would be better

ouwenkg: May be 5s is enough in normal circumstances.

tests/integrate_test/test_block_interval.py
```diff
@@ -0,0 +1,40 @@
...
+    receipt = get_receipt(tx_hash)
+    assert receipt['errorMessage'] == None, "Transaction execution failed"
+
+    metadata = get_metadata()
+    assert metadata['blockInterval'] == 1000000000, "Block interval not changed"
```
leeyr338: 我想确认一下，　这个用例仅为了
１. 测试 blockInterval 的值被设进去了，而不测试
2. 这个 interval 是否生效。

对吗？
如果仅是测试 1 , 这个用例没有问题。
而对于２, 这个用例是没有测试的！

ouwenkg: 这里仅仅是为了测试系统合约里的值是否设置成功，这个 interval 是否生效在外部测试。

leeyr338: okay~

tests/integrate_test/test_block_interval.py
kaikai1024: Please do not use `sleep` to ensure tx execution.

And it's useless, `get_receipt` will retry until get the result.

tests/integrate_test/cita_features_test.sh
```diff
@@ -55,6 +62,11 @@ main() {
     echo -n "5) Check permission denied ... "
     test_perm_denied 2
     echo "Done"
+
+    echo -e "6) Run block interval tests ...\n"
+    test_change_block_interval 2
+    timeout=$(check_height_growth_normal 0 60) || (echo "Block interval changed")
```
kaikai1024: the block_intercal is `1000000000`, How `check height growth normal`?

tests/integrate_test/test_block_interval.py
ouwenkg: Ok

tests/integrate_test/cita_features_test.sh
ouwenkg: This check will failed for the block interval larger moren than 60 and exec echo operation.

kaikai1024: then how to check and ensure it?  Can it(the test case) fail if `height growth normal`?

ouwenkg: Check it again

tests/integrate_test/cita_features_test.sh
```diff
@@ -55,6 +62,13 @@ main() {
     echo -n "5) Check permission denied ... "
     test_perm_denied 2
     echo "Done"
+
+    echo -e "6) Run block interval tests ...\n"
+    test_change_block_interval 2
+    timeout=$(check_height_growth_normal 0 60) && (echo "FAILED"
```
kaikai1024: maybe 10 or 15?

ouwenkg: I want to keep this timeout the same with line 53

kaikai1024: That's ok.

kaikai1024: > I want to keep this timeout the same with line 53

The purpose is not same.

PR #856, opened 2019-09-11
Hard code release version

leeyr338:
Hard code release version. 

## LLT test
- From RPC interface.
Screenshot from 2019-09-11 12-33-05

- From micro-service binary:
Screenshot from 2019-09-11 12-36-06

### DIFF
--- a/cita-auth/build.rs
+++ b/cita-auth/build.rs
@@ -18,7 +18,9 @@ use std::env;
 
 use util::build_info::gen_build_info;
 
+const VERSION: &str = "1.0.0";
+
 fn main() {
     let out_dir = env::var("OUT_DIR").unwrap();
-    gen_build_info(out_dir.as_ref(), "build_info.rs");
+    gen_build_info(out_dir.as_ref(), "build_info.rs", VERSION.to_owned());
 }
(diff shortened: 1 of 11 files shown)


cita-auth/build.rs
```diff
@@ -18,7 +18,9 @@ use std::env;
 
 use util::build_info::gen_build_info;
 
+const VERSION: &str = "1.0.0";
```
kaikai1024: What about put this at a public common type. then other modules use it directly?
```
pub const VERSION
```

leeyr338: Yeah, that's a good option!
But I want to keep a separate version of the evolution capability for each service.

kaikai1024: > But I want to keep a separate version of the evolution capability for each service.

For this situation, just do not use the common version value and define own variable.
What i said can avoid that we miss some variables that shoud be changed to new version.

It's ok for me.  maybe discuss it later.

leeyr338: Okay!
As you say, avoiding miss some variables be changed is a thinking aspect.
I have thought to define the `pub const VERSION` in cita-common, but I don't think this is a good option.

PR #883, opened 2019-10-31
Bug fix: Error estimate for the first calling contract.

leeyr338:
Bug fix: Error estimate for the first calling contract.

 The same transaction will get different result in different state. And the estimate action will change the state, so it should take the most primitive state for each estimate.


cita-executor/core/src/libexecutor/command.rs
```diff
@@ -294,11 +291,18 @@ impl Commander for Executor {
             tx.gas = quota;
             let tx = tx.fake_sign(sender);
 
-            let clone_state = state.clone();
+            // The same transaction will get different result in different state.
```
ouwenkg: Why transaction quota used change in different state ?

leeyr338:
> The bug is use changed state when next exec tx ?

Yes, you got it!

PR #885, opened 2019-11-07
Fix update admin

jerry-yu:
Fix update admin

### DIFF
--- a/scripts/contracts/src/permission_management/Authorization.sol
+++ b/scripts/contracts/src/permission_management/Authorization.sol
@@ -60,7 +60,7 @@ contract Authorization is IAuthorization, ReservedAddrPublic {
             _setAuth(_newAdmin, builtInPermissions[i]);
             // keep send tx and create contract
             if (i != 13 && i!= 14)
-                _cancelAuth(_newAdmin, builtInPermissions[i]);
+                _cancelAuth(oldAdmin, builtInPermissions[i]);
         }
     }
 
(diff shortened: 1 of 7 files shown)


scripts/contracts/src/permission_management/Authorization.sol
```diff
@@ -34,16 +34,36 @@ contract Authorization is IAuthorization, ReservedAddrPublic {
...
+        address oldAdmin = allAccounts[0];
+        if (_newAdmin == oldAdmin) {
+           return;
+        }
+        allAccounts[0] = _newAdmin;
+        allAccounts.push(oldAdmin);
+        for (uint8 i;i < builtInPermissions.length; i++) {
+            _setAuth(_newAdmin, builtInPermissions[i]);
+            // keep send tx and create contract
+            if (i != 13 && i!= 14)
```
leeyr338: what meaning for `13` and `14` ? 
It is better to use a constant define?

jerry-yu: It is not c++

scripts/contracts/src/permission_management/Authorization.sol
```diff
@@ -188,4 +208,15 @@ contract Authorization is IAuthorization, ReservedAddrPublic {
...
+    /// @notice Private: delete the permission to the account
+    function _cancelAuth(address _account, address _permission)
+        private
+        returns (bool)
+    {
+        AddressArray.remove(_account, accounts[_permission]);
+        AddressArray.remove(_permission, permissions[_account]);
+        emit AuthCanceled(_account, _permission);
+        return true;
+    }
```
ouwenkg: https://github.com/cryptape/cita/blob/develop/scripts/contracts/src/permission_management/Authorization.sol#L63, Why not use `_cancelAuth()` directly ?

jerry-yu: has no right

PR #890, opened 2019-11-28
Remove circle resource restrictions

ouwenkg:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,7 +109,6 @@ alias:
     docker:
       - image: cita/cita-build:ubuntu-18.04-20190515
     working_directory: ~/cita-build
-    resource_class: xlarge
 
   - &job-unit-test
     <<: *job-default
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -109,7 +109,6 @@ alias:
     docker:
       - image: cita/cita-build:ubuntu-18.04-20190515
     working_directory: ~/cita-build
-    resource_class: xlarge
```
rainchen: should not remove this line, should comment it out and add a comment for why doing this, for example:

```
# require xlarge class to run all test cases
# resource_class: xlarge
```

ouwenkg: After transfering cita repo to new originazation,  xlarge is not available on free plan in circle ci.

https://circleci.com/docs/2.0/configuration-reference/#resource_class

rainchen: this is an important info, should not remove it, once we use pain plan in future we can have a clue.

PR #922, opened 2020-03-23
support aarch64

rink1969:


Makefile
```diff
@@ -4,11 +4,19 @@ CARGO=RUSTFLAGS='-F warnings -A deprecated' cargo
...
-	scripts/release.sh release
+	scripts/release.sh x86 release
+
+aarch64_debug:
+	$(CARGO) build --all --target aarch64-unknown-linux-gnu
+	scripts/release.sh aarch64 debug
+
+aarch64_release:
+	$(CARGO) build --all  --release --target aarch64-unknown-linux-gnu
+	scripts/release.sh aarch64 release
```
leeyr338, line 19: `aarch64` may has different instruction set.
Is it better to use instruction arch (something like A57, A7) ?

aarch64.md
```diff
@@ -0,0 +1,135 @@
+# 在华为鲲鹏上运行CITA
+
+鲲鹏是华为海思发布的基于ARM架构授权，由华为自主设计完成的鲲鹏芯片。
```
leeyr338: Needs `space` between Chinese and English.

Makefile
rink1969: 指令集是arm v8，只有v8才支持64位。
用这个是因为交叉编译器以及rust对应的target都是用这个开头的。

aarch64.md
rink1969: ok

Makefile
leeyr338: OK

leeyr338 (approved):

